Issue #1, opened 2019-10-22
Adopt CSI for OpenStack extension beginning with Kubernetes 1.19


afritzler:
In order to fix gardener/gardener-extensions#200 and proceed with gardener/gardener-extensions#154 we need to enable the OpenStack CSI in the OpenStack provider.

rfranzke:
The CSI migration is beta with 1.18, though, the CSI plugins for OpenStack are not yet ready to cater with Gardener needs. We will have to contribute there before we can migrate to CSI. Hence, postponing this issue until 1.19.

rfranzke:
I opened https://github.com/kubernetes/cloud-provider-openstack/pull/1021 to allow disabling the individual driver services so that we can run the controller service in the seed and the node service in the shoot. However, the node service is also still depending on/requiring the cloud config and the OpenStack credentials. Anyways, this is already like it is today (the kubelet needs the credentials today as well), so it's at least not getting worse with CSI. https://github.com/kubernetes/cloud-provider-openstack/issues/1020 was opened to track this.

vlerenc:
Wow, despite CSI still. I saw @kayrus opened the ticket. Thanks!

PR #39, opened 2020-03-05
Mount host certificate for KCM/CCM

rfranzke:
The OpenStack extension does now by default mount the host certificates from `/etc/ssl` and `/usr/share/ca-certificates` into the pods that are running in the seed and need to interact with the OpenStack API (kube-apiserver, kube-controller-manager (only until #1 is fixed), and cloud-controller-manager).
This is to accommodate OpenStack environments that are serving certificates signed by CAs not commonly known/trusted.

/cc @bnerd

**Release note**:

```improvement operator
The OpenStack extension does now by default mount the host certificates from `/etc/ssl` and `/usr/share/ca-certificates` into the pods that are running in the seed and need to interact with the OpenStack API (kube-apiserver, kube-controller-manager (only until #1 is fixed), and cloud-controller-manager).
This is to accommodate OpenStack environments that are serving certificates signed by CAs not commonly known/trusted.
```

### DIFF
--- a/charts/internal/cloud-controller-manager/templates/cloud-controller-manager.yaml
+++ b/charts/internal/cloud-controller-manager/templates/cloud-controller-manager.yaml
@@ -117,6 +117,14 @@ spec:
           mountPath: /var/lib/cloud-controller-manager-server
         - name: cloud-provider-config
           mountPath: /etc/kubernetes/cloudprovider
+      # Host certificates are mounted to accommodate OpenStack endpoints that might be served with a certificate
+      # signed by a CA that is not globally trusted.
+        - name: usr-share-ca-certificates
+          mountPath: /usr/share/ca-certificates
+          readOnly: true
+        - name: etc-ssl
+          mountPath: /etc/ssl
+          readOnly: true
       dnsPolicy: ClusterFirst
       restartPolicy: Always
       schedulerName: default-scheduler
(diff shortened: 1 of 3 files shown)


charts/internal/cloud-controller-manager/templates/cloud-controller-manager.yaml
```diff
@@ -117,6 +117,12 @@ spec:
           mountPath: /var/lib/cloud-controller-manager-server
         - name: cloud-provider-config
           mountPath: /etc/kubernetes/cloudprovider
+      # Host certificates are mounted to accommodate OpenStack endpoints that might be served with a certificate
+      # signed by a CA that is not globally trusted.
+        - name: usr-share-ca-certificates
+          mountPath: /usr/share/ca-certificates
+        - name: etc-ssl
+          mountPath: /etc/ssl
```
tim-ebert: Would it make sense to mount them as `readOnly` here, too?
```suggestion
        - name: usr-share-ca-certificates
          mountPath: /usr/share/ca-certificates
          readOnly: true
        - name: etc-ssl
          mountPath: /etc/ssl
          readOnly: true
```

pkg/webhook/controlplane/ensurer.go
```diff
@@ -146,30 +160,20 @@ var (
 
 func ensureVolumeMounts(c *corev1.Container, version string) {
 	c.VolumeMounts = extensionswebhook.EnsureVolumeMountWithName(c.VolumeMounts, cloudProviderConfigKubeControllerManagerVolumeMount)
-
-	if mustMountEtcSSLFolder(version) {
-		c.VolumeMounts = extensionswebhook.EnsureVolumeMountWithName(c.VolumeMounts, etcSSLVolumeMount)
-	}
+	// Host certificates are mounted to accommodate OpenStack endpoints that might be served with a certificate
```
tim-ebert: This will add the volume mounts also to the `kube-apiserver` deployment.
I guess, this is intended, right? Can you then add it to the release note as well?

PR #44, opened 2020-03-13
Validate min workers for autoscaling

timuthy:
This PR enhances the shoot workers validation. It prevents that a worker `minimum` value can be set to 0 if `maximum > 0` since this is not yet supported by autoscaling, see https://github.com/gardener/gardener/pull/2045.

/cc @hardikdr

**Release note**:

```improvement operator
The OpenStack shoot validator now checks that `workers[].minimum != 0` if `workers[].maximum >0` since autoscaling does not support this setup yet.
```


timuthy:
Update: I added the validator to the `Dockerfile` which was missing before.

PR #49, opened 2020-03-21
Use the Metadata server to get the hostname of a VM

vpnachev:
Use the Metadata server to get the hostname of a VM.
This is necessary because different distributions are using different means to set the hostname on the node.

Fixes #

**Release note**:

```improvement operator
The hostname is now retrieved from the metadata server.
```


vpnachev:
/close

The current solution is working reliably on OpenStack.

PR #56, opened 2020-04-01
Add v1.18.0 cloud-controller-manager for 1.18 shoots

rfranzke:
Add v1.18.0 cloud-controller-manager for 1.18 shoots.

Part of gardener/gardener#2095

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -16,3 +16,8 @@ images:
   repository: k8scloudprovider/openstack-cloud-controller-manager
   tag: "v1.17.0"
   targetVersion: ">= 1.15"
+- name: cloud-controller-manager
+  sourceRepository: github.com/kubernetes/cloud-provider-openstack
+  repository: k8scloudprovider/openstack-cloud-controller-manager
+  tag: "v1.18.0"
+  targetVersion: ">= 1.18"


dkistner:
That mean using the openstack out-of-tree ccm beginning k8s v1.18.
@MartinWeindel @mandelsoft Any concerns?

rfranzke:
What do you mean? We are using out-of-tree CCM since 1.15 already.

dkistner:
Ahh sorry my fault...
/lgtm

rfranzke:
@vlerenc why was the `reviewed/lgtm` label not added with https://github.com/gardener/gardener-extension-provider-openstack/pull/56#pullrequestreview-385323804 already?

vlerenc:
@rfranzke Need to check. Since it worked for @dkistner, some problem with @tim-ebert and recognising him as a code owner.

vlerenc:
I need to debug - it should have worked - my bug, I guess.
@dkistner was member of the first mentioned code owner team, while Tim was not.  Tim was only on the second team. That should have worked, but obviously didn't.

rfranzke:
Thanks for looking into it!

vlerenc:
@tim-ebert I can't find a problem other than around the time the configuration was updated, which is very disruptive atm. You are identified as code owner. Can you please create a non-critical label such as `/stale` please?

tim-ebert:
Hmm, worked :D

vlerenc:
OK, then I assume you fell in a gap when the botmanager was updated yesterday around the same time. The problem is that I have GitHub webhook relays in the public internet. They gather events (such as this one from the public GitHub), because it couldn't reach the internal botmanager directly. When it gets redeployed, but the events weren't yet polled from the internal botmanager, its events are lost (only stored in memory). They are not persisted in between the deployments. Sorry for that.

PR #63, opened 2020-04-15
Allow wildcard expressions in FloatingPoolNames

### LINKED ISSUE #62: Drop floating pool name validation
The OpenStack validator checks that the `floatingPoolName` used for shoots is one that was previously defined in the `CloudProfile` by a Gardener operator. However, some OpenStack environments don't manage the floating pools globally (at least not all of them) but sometimes also per project. Hence, such a list in the `CloudProfile` is never complete and we should allow end-users to enter an arbitrary floating pool name in their shoot spec. The list in the `CloudProfile` can still be kept and help users/the dashboard to find out default pools.

timuthy:
This PR adds the support for wildcard expressions when declaring `floatingPools[].name` in the `CloudProfile`.

Fixes #62

`LoadBalancerClasses` in the `ControlPlaneConfig` are now cross checked against the configured `FloatingPoolName` in the `InfrastructureConfig`.

**Release note**:

```improvement operator
It is now possible to use simple wildcard expressions when configuring `floatingPools[].name` in the `CloudProfileConfig`. As a consequence, the `FloatingPoolName` in a shoot cluster is matched against this wildcard expression instead of only allowing a strict match. Please consult the documentation https://github.com/gardener/gardener-extension-provider-openstack/blob/c0588dd370006b9f9fca910037e05387c752654b/docs/usage-as-operator.md for more details.
```

### DIFF
--- a/docs/usage-as-operator.md
+++ b/docs/usage-as-operator.md
@@ -51,6 +51,13 @@ constraints:
 #     floatingSubnetID: "1234"
 #     floatingNetworkID: "4567"
 #     subnetID: "7890"
+# - name: "fp-pool-*"
+#   region: europe
+#   loadBalancerClasses:
+#   - name: lb-class-1
+#     floatingSubnetID: "1234"
+#     floatingNetworkID: "4567"
+#     subnetID: "7890"
   loadBalancerProviders:
   - name: haproxy
 #   region: europe
@@ -59,8 +66,9 @@ constraints:
 ```
 
 Please note that it is possible to configure a region mapping for keystone URLs, floating pools, and load balancer providers.
+Floating pool names may also contains simple wildcard expressions, like `*` or `fp-pool-*` or `*-fp-pool`. Please note that the `*` must be either single or at the beginning or at the end. Consequently, `fp-*-pool` is not possible/allowed.
 The default behavior is that, if found, the regional entry is taken.
-If no entry for the given region exists then the fallback value is the first entry in the list without a `region` field (or the `keystoneURL` value for the keystone URLs).
+If no entry for the given region exists then the fallback value is the most matching entry (w.r.t. wildcard matching) in the list without a `region` field (or the `keystoneURL` value for the keystone URLs).
 Some OpenStack environments don't need these regional mappings, hence, the `region` and `keystoneURLs` fields are optional.
 If your OpenStack environment only has regional values and it doesn't make sense to provide a (non-regional) fallback then simply
 omit `keystoneURL` and always specify `region`.
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/gardener/gardener-extension-provider-openstack
 
-go 1.13
+go 1.14
 
 require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.1.5
--- a/pkg/apis/openstack/validation/controlplane.go
+++ b/pkg/apis/openstack/validation/controlplane.go
@@ -50,14 +50,14 @@ func ValidateControlPlaneConfigUpdate(oldConfig, newConfig *api.ControlPlaneConf
 }
 
 // ValidateControlPlaneConfigAgainstCloudProfile validates the given ControlPlaneConfig against constraints in the given CloudProfile.
-func ValidateControlPlaneConfigAgainstCloudProfile(cpConfig *api.ControlPlaneConfig, shootRegion string, cloudProfile *gardencorev1beta1.CloudProfile, cloudProfileConfig *api.CloudProfileConfig, fldPath *field.Path) field.ErrorList {
+func ValidateControlPlaneConfigAgainstCloudProfile(cpConfig *api.ControlPlaneConfig, shootRegion, floatingPoolName string, cloudProfile *gardencorev1beta1.CloudProfile, cloudProfileConfig *api.CloudProfileConfig, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	if ok, validLoadBalancerProviders := validateLoadBalancerProviderConstraints(cloudProfileConfig.Constraints.LoadBalancerProviders, shootRegion, cpConfig.LoadBalancerProvider); !ok {
 		allErrs = append(allErrs, field.NotSupported(fldPath.Child("loadBalancerProvider"), cpConfig.LoadBalancerProvider, validLoadBalancerProviders))
 	}
 
-	allErrs = append(allErrs, validateLoadBalancerClassesConstraints(cloudProfileConfig.Constraints.FloatingPools, cpConfig.LoadBalancerClasses, shootRegion, fldPath.Child("loadBalancerClasses"))...)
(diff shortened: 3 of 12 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/gardener/gardener-extension-provider-openstack
 
-go 1.13
+go 1.14
```
rfranzke, line 3: Just a reminder: We also have to update the `meta/ci` ref and the pipeline definition there after this PR is merged. Can you take care? :)

pkg/apis/openstack/validation/controlplane_test.go
```diff
@@ -318,11 +320,53 @@ var _ = Describe("ControlPlaneConfig validation", func() {
...
+			})
+
+			controlPlane.LoadBalancerClasses = []api.LoadBalancerClass{lbClasses[1]}
+
+			errorList := ValidateControlPlaneConfigAgainstCloudProfile(controlPlane, region, fpName, cloudProfile, cloudProfileConfig, nilPath)
 
 			Expect(errorList).To(BeEmpty())
 		})
 
+		It("should forbid for a load balancer class which is assigned by a wildcard FloatingPoolName because a more specific one exists", func() {
```
rfranzke, line 349: Hm, not sure if I understand this case, can you elaborate?

pkg/utils/utils.go
```diff
@@ -41,3 +45,21 @@ func SetStringValue(values map[string]interface{}, key string, value *string) {
...
+// Only simple wildcard patterns are supposed to be passed, e.g. '*', 'tex*'.
+func SimpleMatch(pattern, text string) (bool, int) {
+	const wildcard = "*"
+	if pattern == wildcard {
+		return true, 0
+	}
+	if pattern == text {
+		return true, len(text)
+	}
+	if strings.HasSuffix(pattern, wildcard) && strings.HasPrefix(text, pattern[:len(pattern)-1]) {
```
rfranzke, line 60: Hm, this will only work  for `foo-*`, right? What about `*-bar`?

vendor/modules.txt
```diff
@@ -11,10 +11,12 @@ github.com/PuerkitoBio/purell
 # github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578
 github.com/PuerkitoBio/urlesc
 # github.com/ahmetb/gen-crd-api-reference-docs v0.1.5
+## explicit
```
rfranzke, line 14: These changes are related to 1.14?

pkg/apis/openstack/validation/infrastructure.go
```diff
@@ -76,45 +77,63 @@ func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *api.Infrastructure
...
 
 	return allErrs
 }
 
-func validateFloatingPoolNameConstraints(names []api.FloatingPool, region string, name string) (bool, []string) {
+func validateFloatingPoolNameConstraints(fps []api.FloatingPool, region string, name string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if _, err := findFloatingPool(fps, region, name, fldPath.Child("floatingPoolName")); err != nil {
+		allErrs = append(allErrs, err)
+	}
```
rfranzke, line 89:
```suggestion
	_, err := findFloatingPool(fps, region, name, fldPath.Child("floatingPoolName"))
  allErrs = append(allErrs, err)
```

pkg/utils/utils.go
timuthy: The latter one is not supported at the moment. Do you see a necessity of doing this as well?

vendor/modules.txt
timuthy: Correct, it's changed after `make revendor`.

pkg/utils/utils.go
rfranzke: Hm, if it's easy to add then I'd add it right away now. As user I would expect that I can also use `*-bar` expression if `*` and `*-foo` is already supported.

pkg/apis/openstack/validation/controlplane_test.go
timuthy: Sure.

Let's suppose you have this configuration in the cloud profile:

```
...
constraints:
  floatingPools:
  - name: fp-pool-1
    region: europe
    loadBalancerClasses:
    - name: lb-class-1
      floatingSubnetID: "1234"
      floatingNetworkID: "4567"
      subnetID: "7890"
  - name: "*"
    region: europe
    loadBalancerClasses:
    - name: lb-class-2
      floatingSubnetID: "1234"
      floatingNetworkID: "4567"
      subnetID: "7890"
```

If the shoot is configured for FP `fp-pool-1` it must not be allowed to let it use LB class `lb-class-2` because there is a closer match in the cloud profile that only specifies `lb-class-1`.

Let me revise the naming of this test but do you generally agree with this use case?

go.mod
timuthy: Absolutely! Thanks for letting me know.

pkg/apis/openstack/validation/controlplane_test.go
rfranzke: Ah, understood now, yep, thanks!

vendor/modules.txt
rfranzke: Thanks!

pkg/apis/openstack/validation/infrastructure.go
timuthy: I think it's not desirable to let `allErrs` contain a `nil` element. I changed the `findFloatingPool` to return an `ErrorList` instead.

timuthy:
Thanks for the review @rfranzke. I have addressed your issues. Can you take another look?

pkg/apis/openstack/validation/controlplane.go
```diff
@@ -121,55 +121,37 @@ func validateZoneConstraints(regions []gardencorev1beta1.Region, region, zone st
...
+		allErrs = append(allErrs, field.Invalid(field.NewPath("floatingPoolName"), floatingPoolName, errs.ToAggregate().Error()))
 		return allErrs
 	}
 
-	for i, lbClass := range lbClasses {
-		var (
-			valid       bool
-			validValues []string
-		)
+	if len(fp.LoadBalancerClasses) > 0 {
```
rfranzke, line 136: What about exiting early here to remove the levels of nesting?

docs/usage-as-operator.md
```diff
@@ -59,8 +66,9 @@ constraints:
 ```
 
 Please note that it is possible to configure a region mapping for keystone URLs, floating pools, and load balancer providers.
+Floating pool names may also contains simple wildcard expressions, like "*" or "fp-pool-*".
```
rfranzke, line 69:
```suggestion
Floating pool names may also contains simple wildcard expressions, like `*` or `fp-pool-*` or `*-fp-pool`. Please note that the `*` must be either single or at the beginning or at the end. Consequently, `fp-*-pool` is not possible/allowed.
```

timuthy:
Ready for another round @rfranzke :)

PR #65, opened 2020-04-17
CSI migration beginning with 1.19, convert cloud-provider configmap into secret, remove zone from ControlPlaneConfig

### LINKED ISSUE #1: Adopt CSI for OpenStack extension beginning with Kubernetes 1.19

### LINKED ISSUE #61: Remove zone from ControlPlaneConfig
The `zone` field should be removed from the `ControlPlaneConfig`. It was taken over during extensibility from the in-tree behaviour, but it is not needed. In fact, it only influences the `zone` parameter in the default `StorageClass` that gets installed to the shoot cluster, though, this does not work well with multi-zone shoots. The default behaviour of the Cinder volume plugin - if no `zone` parameter is specified - is to randomize the zones.

**Why is this needed**:
Better default configuration for OpenStack shoots.

rfranzke:
This PR adds support for CSI (+ migration for existing clusters).
CSI will be used for 1.19 shoots. Older shoots will keep using the in-tree legacy volume provisioner.
Also, it converts the cloud-provider `ConfigMap` to a `Secret`.
Also, it removes the `zone` field from the `ControlPlaneConfig`.

Fixes #1 
Fixes #61

OpenStack CSI migration is still alpha with 1.18 but will be promoted to beta with 1.19. When we work on 1.19 support for OpenStack we should revisit the CSI and probably update the image versions.

**Release note**:

```noteworthy user
The `zone` field in the `ControlPlaneConfig` has been deprecated and will be removed in a future version. It is no longer required to set and has no effect anymore. Consider updating your deployment manifests.
```
```noteworthy operator
The `cloud-provider-config-cloud-controller-manager` has been renamed to `cloud-provider-config` and is now a `Secret` instead of a `ConfigMap`. The `cloud-provider-config-kube-controller-manager` has been renamed to `cloud-provider-disk-config` and is now a `Secret` instead of a `ConfigMap`.
```


ialidzhikov:
> I will play with it asap.

I played with pvcs and volumesnapshots, it worked well.

rfranzke:
Thanks @ialidzhikov!
Anybody else planning to review it? If not, I want to merge it tomorrow.

rfranzke:
@tim-ebert does it mean you won't review it or whether you'd want me to wait? 😄

tim-ebert:
Sorry, that was ambiguous. 
I won’t review😅

gardener-botmanager:
@rfranzke Command /merge is not known.

PR #76, opened 2020-05-14
add domain field to floatingpool and validate it

### LINKED ISSUE #73: Add domain field to `openstackv1alpha1.FloatingPool`
Let's add a new optional `domain *string` field to the [`openstackv1alpha1.FloatingPool`](https://github.com/gardener/gardener-extension-provider-openstack/blob/master/pkg/apis/openstack/v1alpha1/types_cloudprofile.go#L63-L72). It allows to configure that the given floating pool is only available in the specified domain. If no domain is specified then it is valid in all domains:

```yaml
apiVersion: core.gardener.cloud/v1beta1
kind: CloudProfile
metadata:
  name: openstack
spec:
  type: openstack
  ...
  providerConfig:
    kind: CloudProfileConfig
    ...
    constraints:
      floatingPools:
      - name: fp-pool-1 #valid in `domain1` domain in `eu-1` region
        region: eu-1
        domain: domain1
      - name: fp-pool-2 # valid in all domains in `eu-1` region
        region: eu-1
      - name: fp-pool-3 # valid in all domains and all regions
      loadBalancerProviders:
      - name: haproxy
```

**Why is this needed**:
In some OpenStack environments the floating pools differ based on region (already supported) **and domain** (not yet supported). Reflecting this in the `CloudProfile` will make the user experience better and allows the Gardener Dashboard to provide the correct values in the cluster creation dialog.

MartinWeindel:
Add an optional domain field to FloatingPool to allow configuring that the given floating pool is only available in the specified domain.
The cloud profile validation now checks the specified floatingpools for consistent combinations of name/region/domain.
The shoot validation prevents using floating pools from a different region/domain combination.

Example:

```yaml
apiVersion: core.gardener.cloud/v1beta1
kind: CloudProfile
metadata:
  name: openstack
spec:
  type: openstack
  ...
  providerConfig:
    kind: CloudProfileConfig
    ...
    constraints:
      floatingPools:
      - name: fp-pool-1 #valid in `domain1` domain in `eu-1` region
        region: eu-1
        domain: domain1
      - name: fp-pool-2 # valid in all domains in `eu-1` region (but not for domain `domain1`)
        region: eu-1
      - name: fp-pool-3 # valid in all domains and all regions (but not for region `eu-1`)
      loadBalancerProviders:
      - name: haproxy
```

Fixes #73

**Release note**:

```improvement operator
added optional domain field to floatingpools in the cloud profile constraints
```

### DIFF
--- a/pkg/apis/openstack/v1alpha1/types_cloudprofile.go
+++ b/pkg/apis/openstack/v1alpha1/types_cloudprofile.go
@@ -66,6 +66,12 @@ type FloatingPool struct {
 	// Region is the region name.
 	// +optional
 	Region *string `json:"region,omitempty"`
+	// Domain is the domain name.
+	// +optional
+	Domain *string `json:"domain,omitempty"`
+	// NonConstraining specifies whether this floating pool is not constraining, that means additionally available independent of other FP constraints.
+	// +optional
+	NonConstraining *bool `json:"nonConstraining,omitempty"`
 	// LoadBalancerClasses contains a list of supported labeled load balancer network settings.
 	// +optional
 	LoadBalancerClasses []LoadBalancerClass `json:"loadBalancerClasses,omitempty"`
--- a/pkg/apis/openstack/validation/cloudprofile.go
+++ b/pkg/apis/openstack/validation/cloudprofile.go
@@ -19,6 +19,8 @@ import (
 	"net"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	api "github.com/gardener/gardener-extension-provider-openstack/pkg/apis/openstack"
 
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -33,21 +35,34 @@ func ValidateCloudProfileConfig(cloudProfile *api.CloudProfileConfig) field.Erro
 		allErrs = append(allErrs, field.Required(floatingPoolPath, "must provide at least one floating pool"))
 	}
 
-	regionsFound := map[string]struct{}{}
+	combinationFound := sets.NewString()
 	for i, pool := range cloudProfile.Constraints.FloatingPools {
 		idxPath := floatingPoolPath.Index(i)
 		if len(pool.Name) == 0 {
 			allErrs = append(allErrs, field.Required(idxPath.Child("name"), "must provide a name"))
 		}
 
-		if pool.Region != nil {
-			if len(*pool.Region) == 0 {
-				allErrs = append(allErrs, field.Required(idxPath.Child("region"), "must provide a region if key is present"))
+		if pool.Region != nil || pool.Domain != nil {
+			region := "*"
+			domain := "*"
+			if pool.Region != nil {
+				if len(*pool.Region) == 0 {
+					allErrs = append(allErrs, field.Required(idxPath.Child("region"), "must provide a region if key is present"))
+				}
+				region = *pool.Region
 			}
-			if _, ok := regionsFound[*pool.Region]; ok {
-				allErrs = append(allErrs, field.Duplicate(idxPath.Child("region"), *pool.Region))
+			if pool.Domain != nil {
+				if len(*pool.Domain) == 0 {
+					allErrs = append(allErrs, field.Required(idxPath.Child("domain"), "must provide a domain if key is present"))
+				}
+				domain = *pool.Domain
 			}
-			regionsFound[*pool.Region] = struct{}{}
+			key := fmt.Sprintf("%s,%s,%s", pool.Name, domain, region)
+			if combinationFound.Has(key) {
+				// duplicate for given name/domain/region combination
+				allErrs = append(allErrs, field.Duplicate(idxPath.Child("name"), pool.Name))
+			}
+			combinationFound.Insert(key)
 		}
 	}
 
@@ -56,7 +71,7 @@ func ValidateCloudProfileConfig(cloudProfile *api.CloudProfileConfig) field.Erro
 		allErrs = append(allErrs, field.Required(loadBalancerProviderPath, "must provide at least one load balancer provider"))
 	}
 
-	regionsFound = map[string]struct{}{}
+	regionsFound := sets.NewString()
 	for i, pool := range cloudProfile.Constraints.LoadBalancerProviders {
 		idxPath := loadBalancerProviderPath.Index(i)
 
@@ -68,10 +83,10 @@ func ValidateCloudProfileConfig(cloudProfile *api.CloudProfileConfig) field.Erro
 			if len(*pool.Region) == 0 {
 				allErrs = append(allErrs, field.Required(idxPath.Child("region"), "must provide a region if key is present"))
 			}
-			if _, ok := regionsFound[*pool.Region]; ok {
+			if regionsFound.Has(*pool.Region) {
 				allErrs = append(allErrs, field.Duplicate(idxPath.Child("region"), *pool.Region))
 			}
-			regionsFound[*pool.Region] = struct{}{}
+			regionsFound.Insert(*pool.Region)
 		}
 	}
 
@@ -105,7 +120,7 @@ func ValidateCloudProfileConfig(cloudProfile *api.CloudProfileConfig) field.Erro
 		allErrs = append(allErrs, field.Required(field.NewPath("keyStoneURL"), "must provide the URL to KeyStone"))
 	}
 
-	regionsFound = map[string]struct{}{}
+	regionsFound = sets.NewString()
 	for i, val := range cloudProfile.KeyStoneURLs {
 		idxPath := field.NewPath("keyStoneURLs").Index(i)
 
@@ -117,10 +132,10 @@ func ValidateCloudProfileConfig(cloudProfile *api.CloudProfileConfig) field.Erro
 			allErrs = append(allErrs, field.Required(idxPath.Child("url"), "must provide an url"))
 		}
 
-		if _, ok := regionsFound[val.Region]; ok {
+		if regionsFound.Has(val.Region) {
 			allErrs = append(allErrs, field.Duplicate(idxPath.Child("region"), val.Region))
 		}
-		regionsFound[val.Region] = struct{}{}
+		regionsFound.Insert(val.Region)
 	}
 
 	for i, ip := range cloudProfile.DNSServers {
--- a/pkg/apis/openstack/validation/infrastructure.go
+++ b/pkg/apis/openstack/validation/infrastructure.go
@@ -15,11 +15,15 @@
 package validation
 
 import (
+	"reflect"
+	"sort"
+
 	api "github.com/gardener/gardener-extension-provider-openstack/pkg/apis/openstack"
 	"github.com/gardener/gardener-extension-provider-openstack/pkg/utils"
 
 	cidrvalidation "github.com/gardener/gardener/pkg/utils/validation/cidr"
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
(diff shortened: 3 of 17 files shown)


pkg/apis/openstack/validation/cloudprofile.go
```diff
@@ -33,21 +33,33 @@ func ValidateCloudProfileConfig(cloudProfile *api.CloudProfileConfig) field.Erro
...
+			if pool.Domain != nil {
+				if len(*pool.Domain) == 0 {
+					allErrs = append(allErrs, field.Required(idxPath.Child("domain"), "must provide a domain if key is present"))
+				}
+				domain = *pool.Domain
 			}
-			regionsFound[*pool.Region] = struct{}{}
+			key := fmt.Sprintf("%s,%s,%s", pool.Name, domain, region)
+			if _, ok := regionsAndDomainsFound[key]; ok {
+				allErrs = append(allErrs, field.Invalid(idxPath.Child("name"), pool.Name, "duplicate for given name/domain/region combination"))
```
rfranzke, line 60:
```suggestion
				allErrs = append(allErrs, field.Duplicate(idxPath.Child("name"), pool.Name))
```

pkg/apis/openstack/validation/infrastructure.go
```diff
@@ -74,66 +77,71 @@ func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *api.Infrastructure
...
+func validateFloatingPoolNameConstraints(fps []api.FloatingPool, domain, region string, name string, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
-	_, errs := findFloatingPool(fps, region, name, fldPath.Child("floatingPoolName"))
+	_, errs := findFloatingPool(fps, domain, region, name, fldPath.Child("floatingPoolName"))
 	allErrs = append(allErrs, errs...)
 	return allErrs
 }
 
-func findFloatingPool(floatingPools []api.FloatingPool, shootRegion, floatingPoolName string, fldPath *field.Path) (*api.FloatingPool, field.ErrorList) {
+func findFloatingPool(floatingPools []api.FloatingPool, domain, shootRegion, floatingPoolName string, fldPath *field.Path) (*api.FloatingPool, field.ErrorList) {
```
rfranzke, line 95: /cc @timuthy
can you check this function please?

pkg/apis/openstack/validation/infrastructure_test.go
```diff
@@ -196,15 +197,30 @@ var _ = Describe("InfrastructureConfig validation", func() {
...
+
+			errorList := ValidateInfrastructureConfigAgainstCloudProfile(infrastructureConfig, domain, region, cloudProfileConfig, nilPath)
+
+			Expect(errorList).To(ConsistOfFields(Fields{
+				"Type":  Equal(field.ErrorTypeNotSupported),
+				"Field": Equal("floatingPoolName"),
+			}))
+		})
+
+		It("should always allow using the non-regional floating pool name", func() {
```
rfranzke, line 222: Hmm, but if the floating pool has a domain then it should only be allowed if the domain matches, or?

pkg/internal/credentials.go
```diff
@@ -18,8 +18,10 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/gardener/gardener-extension-provider-openstack/pkg/openstack"
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
+	"github.com/gardener/gardener/pkg/apis/core/v1beta1"
```
rfranzke, line 22:
```suggestion
	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
```

pkg/validator/shoot_handler.go
```diff
@@ -50,9 +51,16 @@ func (v *Shoot) Handle(ctx context.Context, req admission.Request) admission.Res
 		return admission.Allowed("webhook not responsible for this provider")
 	}
 
+	// Get credentials
+	credentials, err := internal.GetCredentialsBySecretBinding(ctx, v.client, client.ObjectKey{Namespace: shoot.Namespace, Name: shoot.Spec.SecretBindingName})
+	if err != nil {
+		v.Logger.Error(err, "could not get service account from SecretBindingName %s", shoot.Spec.SecretBindingName)
```
rfranzke, line 57:
```suggestion
		v.Logger.Error(err, "could not get credentials from SecretBindingName %s", shoot.Spec.SecretBindingName)
```

rfranzke:
/kind api-change
/area usability
/needs second-opinion
/size/m

pkg/apis/openstack/validation/infrastructure_test.go
MartinWeindel: yes, see test "should only allow domain specific floating pool name for this domain"
I will rename this to "should always allow using a global floating pool name in presence of a regional fp"

rfranzke: +1

pkg/apis/openstack/validation/infrastructure.go
```diff
@@ -74,66 +77,71 @@ func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *api.Infrastructure
...
+}
+
+func findFloatingPoolSub(floatingPools []api.FloatingPool, domain, shootRegion *string, floatingPoolName string, validValues map[string]struct{}) *api.FloatingPool {
+	var (
+		candidate      *api.FloatingPool
+		candidateScore int
+	)
+
+	for _, fp := range floatingPools {
+		if reflect.DeepEqual(domain, fp.Domain) && reflect.DeepEqual(shootRegion, fp.Region) {
```
timuthy, line 134: This introduces a change to the general validation logic. Previously, we forbade to use a non-regional FloatingPool if the CloudProfile specified FloatingPool(s) for the region used by the shoot.
A regional configuration has not been part of FloatingPools since day one. Hence, the assumption is that a regional setting for FloatingPools overrules a non regional setting.

I'm not sure if we proactively want to open this restriction. @rfranzke WDYT?

MartinWeindel: I have added a table test with all  relevant cases to make reasoning easier. See pkg/apis/openstack/validation/infrastructure_test.go `Describe("#FindFloatingPool")...`

timuthy: Okay, but what is the motivation to open up the mentioned restriction?

MartinWeindel: With the last push the logic has been reverted to the original behaviour. Regional or domain settings overrule any non-regional setting.

MartinWeindel:
@rfranzke @timuthy With Uwe I discussed the introduction of an `additional`field to change the logic for a floating pool from constraining to additive. These changes are in the second commit. We need to decide if this addition should be kept to allow more flexibility.

Example:

```yaml
apiVersion: core.gardener.cloud/v1beta1
kind: CloudProfile
metadata:
  name: openstack
spec:
  type: openstack
  ...
  providerConfig:
    kind: CloudProfileConfig
    ...
    constraints:
      floatingPools:
      - name: fp-pool-1 #valid in `domain1` domain and `eu-1` region
        region: eu-1
        domain: domain1
      - name: fp-pool-2 # valid in all domains in `eu-1` region (additionally even in domain `domain1` as an alternative to `fp-pool-1`)
        region: eu-1
        additional: true
      - name: fp-pool-3 # valid in all domains and all regions (but not for region `eu-1`)
      loadBalancerProviders:
      - name: haproxy
```

rfranzke:
@MartinWeindel thanks, makes sense. I'd just try to find a better name for `additional` (it's not too clear what its purpose is without looking deeper into the structure of `floatingPoolNames`).

MartinWeindel:
> @MartinWeindel thanks, makes sense. I'd just try to find a better name for `additional` (it's not too clear what its purpose is without looking deeper into the structure of `floatingPoolNames`).

@rfranzke how about `non-constraining` ?

rfranzke:
Yes, sounds good, thanks. Just use camelCase instead of kebab-case. :)

pkg/apis/openstack/v1alpha1/types_cloudprofile.go
```diff
@@ -66,6 +66,12 @@ type FloatingPool struct {
 	// Region is the region name.
 	// +optional
 	Region *string `json:"region,omitempty"`
+	// Domain is the domain name.
+	// +optional
+	Domain *string `json:"domain,omitempty"`
+	// NonConstraining specifies whether this floating pool is not constraining, that means additionally available independent of other FP constraints.
+	// +optional
+	NonConstraining *bool `json:"additional,omitempty"`
```
rfranzke, line 74:
```suggestion
	NonConstraining *bool `json:"nonConstraining,omitempty"`
```

pkg/apis/openstack/validation/infrastructure.go
```diff
@@ -74,66 +78,104 @@ func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *api.Infrastructure
...
+	}
+	for key := range additionalValidValues {
+		validValuesList = append(validValuesList, key)
+	}
+	sort.Strings(validValuesList)
+	allErrs = append(allErrs, field.NotSupported(fldPath, floatingPoolName, validValuesList))
+	return nil, allErrs
+}
+
+func findFloatingPoolSub(floatingPools []api.FloatingPool, domain, shootRegion *string, floatingPoolName string,
```
rfranzke, line 153: Can we find a better function name? What is "sub"?

pkg/apis/openstack/validation/infrastructure.go
```diff
@@ -74,66 +78,104 @@ func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *api.Infrastructure
...
 	allErrs := field.ErrorList{}
-	_, errs := findFloatingPool(fps, region, name, fldPath.Child("floatingPoolName"))
+	_, errs := FindFloatingPool(fps, domain, region, name, fldPath.Child("floatingPoolName"))
 	allErrs = append(allErrs, errs...)
 	return allErrs
 }
 
-func findFloatingPool(floatingPools []api.FloatingPool, shootRegion, floatingPoolName string, fldPath *field.Path) (*api.FloatingPool, field.ErrorList) {
+// FindFloatingPool finds best match for given domain, region, and name
+func FindFloatingPool(floatingPools []api.FloatingPool, domain, shootRegion, floatingPoolName string, fldPath *field.Path) (*api.FloatingPool, field.ErrorList) {
```
rfranzke, line 97: @timuthy could you check this as well? You're most familiar with the code, I guess...

pkg/apis/openstack/v1alpha1/types_cloudprofile.go
MartinWeindel: thanks

pkg/apis/openstack/validation/infrastructure.go
MartinWeindel: renamed to findFloatingPoolCandidate

pkg/apis/openstack/validation/infrastructure.go
```diff
@@ -74,66 +78,105 @@ func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *api.Infrastructure
...
+		return foundRegion, allErrs
+	}
+	if foundDomain != nil && len(validValuesRegion) == 0 {
+		return foundDomain, allErrs
+	}
+	if len(validValuesRegion) != 0 && len(validValuesDomain) != 0 {
+		for value := range validValuesRegion {
+			if validValuesDomain.Has(value) {
+				validValues.Insert(value)
 			}
```
timuthy, line 127: IIUC, this can only happen if the same FP is configured twice which is at the same time forbidden by the cloud profile validation logic you added in `pkg/apis/openstack/validation/cloudprofile.go`. 
Seems like unreachable code?
I'm also not sure why we don't allow the union of `validValuesDomain` and `validValuesRegion`. Can you elaborate?

MartinWeindel: This was also my first thought. But the union of `validValuesDomain`and `validValuesRegion` is not quite correct, as constraints of both must be fulfilled.
It must be the intersection of both, but only if both are specified.
The validation logic in the cloud profile only forbids duplicates of the tuple `(name, region, domain)`.

Example:
```yaml
   constraints:
      floatingPools:
      - name: fp1
        region: eu
      - name: fp1
        domain: dom1
      - name: fp2
        region: us
```

Case 1: region `eu`, domain `dom1`
validValuesRegion=`["fp1"]`, validValuesDomain=`["fp1"]` => validValues is the intersection, i.e. `["fp1"]`

Case 2: region `us`, domain `dom1`
validValuesRegion=`["fp2"]`, validValuesDomain=`["fp1"]`,  => validValues is the intersection, i.e. `[]`.

timuthy: I see, thanks for explaining. Can you add a unit test for this case as well?

Also, you can make use `validValuesDomain.Intersection(validValuesRegion)` instead of looping and adding the intersected values separately.

MartinWeindel: This is already fully covered by the "FindFloatingPool table" test (see https://github.com/gardener/gardener-extension-provider-openstack/pull/76/files#diff-aa42ff40241b80a4ca184b047877b3faR460-R462)

MartinWeindel:
commits have been squashed to a single one

pkg/apis/openstack/validation/infrastructure.go
timuthy: Don't we miss a case for which the `validValues > 0` after the intersection? At least all tests pass when I add the following panic to `FindFloatingPool`.

	if len(validValuesRegion) != 0 && len(validValuesDomain) != 0 {
		// if region and domain are constrainted separately, only values are valid which are valid for both
		validValues = validValuesRegion.Intersection(validValuesDomain)
		if validValues.Len() > 0 {
			panic("covered")
		}

rfranzke:
/status author-action
@MartinWeindel

pkg/apis/openstack/validation/infrastructure.go
MartinWeindel: Ok, now also covered in the case an invalid fp is requested (see last commit)

PR #92, opened 2020-06-09
Allow to configure subnet which a router should be attached to

dkistner:
Co-authored-by: Raphael Vogel <raphael.vogel@sap.com>
Co-authored-by: Martin Weindel <martin.weindel@sap.com>

/kind enhancement
/priority normal
/platform openstack

Allow to configure to which flooting pool subnet the router should be attached to.
See [here](https://www.terraform.io/docs/providers/openstack/r/networking_router_v2.html#external_fixed_ip).

~~Be aware there is bug in the Neutron api which does not allow updating the router attached to an subnet. There is already a [fix](https://bugs.launchpad.net/neutron/+bug/1812118), but might be not available on all openstack installations.~~

**Update**: The issue which lead to failed tf re-applies in not cause by the Neutron bug, but by a bug in the OpenStack Terraform provider. A fix is available and we moved already to a terraform provider version which include this fix. See #103 

**Release note**:

```improvement user
The floating pool subnet which a router should be attached to can now be specified.
```


charts/internal/openstack-infra/templates/main.tf
```diff
@@ -21,6 +21,11 @@ resource "openstack_networking_router_v2" "router" {
   name                = "{{ required "clusterName is required" .Values.clusterName }}"
   region              = "{{ required "openstack.region is required" .Values.openstack.region }}"
   external_network_id = "${data.openstack_networking_network_v2.fip.id}"
+  {{ if and .Values.router.floatingPoolSubnetID -}}
```
MartinWeindel, line 24: `and` should be dropped

example/30-infrastructure.yaml
```diff
@@ -62,5 +62,6 @@ spec:
     networks:
     # router:
     #   id: 1234
+    #   floatingPoolSubnetId: my-floating-pool-subnet-id
```
rfranzke, line 65: Why didn't you put it next to `floatingPoolName` but below `router`?

pkg/internal/infrastructure/terraform.go
```diff
@@ -66,12 +66,19 @@ func ComputeTerraformerChartValues(
...
 
-	if router := config.Networks.Router; router != nil {
-		createRouter = false
-		routerID = router.ID
+	if config.Networks.Router != nil {
+		if config.Networks.Router.ID != nil {
+			createRouter = false
+			routerConfig["id"] = *config.Networks.Router.ID
+		}
+		if createRouter && config.Networks.Router.FloatingPoolSubnetID != nil {
```
rfranzke, line 79: Can you add a unit test for this case?

dkistner:
We saw the issue which lead to failed re run of the terraform job is actually not caused by the neutron api issue but rather by a bug in the OpenStack Terraform provider. Our used terraform was too old and does not include the fix. 

I created a PR on the terraformer repo to bump the OpenStack provider version. When this merged we can consider to bring in this change. https://github.com/gardener/terraformer/pull/41

Another question which come into my mind. Is there any scenario when a router should be again detached from the fip subnet? If not this field need to be immutable. 

cc @MartinWeindel, @rfranzke, @RaphaelVogel

dkistner:
We have now the required Terraform version in. I will remove the draft state of the PR now.

@MartinWeindel, @mandelsoft, @rfranzke Do you have an option on that?
> Is there any scenario where a router should be again detached from the fip subnet again? If not this field need to be immutable.

pkg/apis/openstack/validation/infrastructure_test.go
```diff
@@ -45,7 +45,7 @@ var _ = Describe("InfrastructureConfig validation", func() {
 			FloatingPoolName: floatingPoolName1,
 			Networks: api.Networks{
 				Router: &api.Router{
-					ID: "hugo",
+					ID: makeStringPointer("hugo"),
```
rfranzke, line 48: Nit: You could use `k8s.io/pointer` similar to #100

example/30-infrastructure.yaml
rfranzke: Hm, but the `floatingPoolName` is also a property of the router. https://github.com/gardener/gardener-extension-provider-openstack/pull/92/files#diff-f8129006b7b8dad2ed592a618872d49fR23
Also, I find it confusing that we use "name" for the floating pool but "id" for the floating pool subnet.

rfranzke:
https://github.com/gardener/gardener-extension-provider-openstack/blob/master/docs/usage-as-end-user.md#infrastructureconfig should be enhanced

example/30-infrastructure.yaml
dkistner: Having the name instead of the idea is a great proposal and we can fetch the corresponding id probably via a data source. I'll do that.

For the structure of the `router` and the `floatingPoolName` I struggle a bit. IMO both belongs to the `router` section, but moving the `floatingPoolName` field to the `router` section would require some migration logic. So what would you suggest? Moving the the `subnetName` to same level as the `floatingPoolName` or the other way around? 

cc @MartinWeindel

rfranzke: I would probably add the `floatingPoolSubnetName` to the same level as `floatingPoolName` now, and strive for an improved structure by moving both to `router` in the next API version. But I have no strong opinion. If you think it's okay as it's now then fine.

dkistner: Sound like a great proposal. I will adapt the PR accordingly and open an issue to restructure the `router` section in the infraconfig for next api-version.

dkistner:
I have addressed the requested changes, added some documentation and squashed the commits of this pr into one commit.

dkistner:
I have added the missing tests and did some DRY in `pkg/internal/infrastructure/terraform_test.go`

dkistner:
@MartinWeindel As a few things have been changes. Could you please have another look?

PR #97, opened 2020-06-16
Cleanup var references

### LINKED ISSUE #96: Adapt to terraform v0.12 language changes

/area open-source
/kind cleanup
/priority normal
/platform openstack

provider-openstack needs an adaptation of the terraform configuration to v0.12. For provider-aws this is done with this PR - https://github.com/gardener/gardener-extension-provider-aws/pull/111. 

**Why is this needed**:
Currently the terraformer run is only omitting warnings but in a future version of terraform, the warnings will be switched to errors.

afritzler:
/kind cleanup
/priority normal
/platform openstack

Remove warnings for Terraform v0.12.x runs.

Fixes #96 

**Release note**:
```improvement operator
Until now `provider-openstack` was maintaining a Terraform configuration which is both `v0.12` and `v0.11` compatible. The Terraform configuration is now adapted to the new Terraform language which makes it Terraform `v0.11` incompatible.
```


charts/internal/openstack-infra/templates/main.tf
```diff
@@ -43,7 +43,7 @@ resource "openstack_networking_subnet_v2" "cluster" {
 
 resource "openstack_networking_router_interface_v2" "router_nodes" {
   router_id = "{{ required "router.id is required" $.Values.router.id }}"
```
ialidzhikov, line 45: I think this PR is not fixing all of the usage of the old syntax. For example 
```
  router_id = "{{ required "router.id is required" $.Values.router.id }}"
```

can be 
```
  router_id = "${openstack_networking_router_v2.router.id}"
```

which is still deprecated usage.

afritzler: Good point! The `DefaultRouterID` should be fixed now by 
```
DefaultRouterID = "openstack_networking_router_v2.router.id"
```

ialidzhikov: Hmm, still:
```
  router_id = "openstack_networking_router_v2.router.id"
```

!=


```
  router_id = openstack_networking_router_v2.router.id
```

We need the quotes only when and existing ID is specified. We don't need the quotes when we use the tf expresison.

afritzler: Actually we don't need the quotes in the rendered helm output at all.

ialidzhikov: When routerID is specified, you can pass the chart value explicitly quoted. At least this was my approach in the respective PR for provider-aws.

rfranzke:
@afritzler can you please update the PR?

afritzler:
Done.

charts/internal/openstack-infra/templates/main.tf
afritzler: Not sure if I am missing something but this is the same approach as in your aws provider PR https://github.com/ialidzhikov/gardener-extension-provider-aws/blob/25b75a6ae4c08a850c04d68c6a8516fbbb6deb12/pkg/controller/infrastructure/actuator_reconcile.go#L111

charts/internal/openstack-infra/templates/main.tf
```diff
@@ -3,8 +3,8 @@ provider "openstack" {
   domain_name = "{{ required "openstack.domainName is required" .Values.openstack.domainName }}"
   tenant_name = "{{ required "openstack.tenantName is required" .Values.openstack.tenantName }}"
   region      = "{{ required "openstack.region is required" .Values.openstack.region }}"
-  user_name   = "${var.USER_NAME}"
-  password    = "${var.PASSWORD}"
+  user_name   = var.USER_NAME
+  password    = var.PASSWORD
```
ialidzhikov, line 7: Can you adapt also `variables.tf`?

charts/internal/openstack-infra/templates/main.tf
ialidzhikov:
```
  router_id = "openstack_networking_router_v2.router.id"
```

should fail that there is no existing router with this id.

ialidzhikov:
@afritzler , assuming that you are also busy with other things, I opened #112 where I applied also the changes requested in this PR and added you as co-author. I hope that this is fine for your.

PR #160, opened 2020-10-09
added nodeVolumeAttachLimit

mganter:

/area robustness
/area storage
/kind enhancement
/priority normal
/platform openstack

The introduced flag tells the CSI-Driver, that there is a upper limit for mounted volumes on a node. 

**Release note**:

```improvement operator
Added possibility to set nodeVolumeAttachLimit within the cloud profile.
```

### DIFF
--- a/charts/internal/cloud-provider-config/templates/cloud-provider-disk-config-csi.yaml
+++ b/charts/internal/cloud-provider-config/templates/cloud-provider-disk-config-csi.yaml
@@ -5,6 +5,9 @@
 
 [BlockStorage]
 rescan-on-resize={{ .Values.rescanBlockStorageOnResize }}
+{{- if .Values.nodeVolumeAttachLimit }}
+node-volume-attach-limit={{ .Values.nodeVolumeAttachLimit }}
+{{- end -}}
 {{- end -}}
 {{- if semverCompare ">= 1.19-0" .Values.kubernetesVersion }}
 ---
--- a/charts/internal/cloud-provider-config/values.yaml
+++ b/charts/internal/cloud-provider-config/values.yaml
@@ -15,6 +15,7 @@ dhcpDomain: foobar
 requestTimeout: 2s
 useOctavia: false
 rescanBlockStorageOnResize: false
+# nodeVolumeAttachLimit: 25
 # floatingClasses:
 # - name: A
 #   floatingNetworkID: "1234"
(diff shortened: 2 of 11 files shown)


gardener-robot-ci-3:
Thank you @mganter for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

dkistner:
Hi @mganter.
Thanks you very much for the PR. 
I will have a look soon.

dkistner (commented):
Thanks @mganter 
I had a brief look. Can you elaborate a little bit on the reasons why the amount of volumes per node need to be limited?
Should that not rather be configured for each `Shoot` cluster individually instead of globally via the `CloudProfile`?

mganter:
The limit exists inside of OpenStack. With this option you can tell the scheduler, that he shall not schedule more cinder volumes to a specific node than the specified amount. 

If the scheduler does't know it, he will schedule pods to nodes where they can't be satisfied in terms of volumes. So pods will remain in container creation state.

dkistner (changes requested):
Thanks for explanation. I played a little bit around with it and configured a limit of just two disks per node.
If I try to schedule more pods with disk to a node the schedules tells:
```
I1019 19:50:13.426597       1 factory.go:445] "Unable to schedule pod; no fit; waiting" pod="default/some-persistence-2" err="0/1 nodes are available: 1 node(s) exceed max volume count."
```
Also there is an event which indicate the same:
```
2m34s       Warning   FailedScheduling          pod/some-persistence-2                                            0/1 nodes are available: 1 node(s) exceed max volume count.
```

The pr looks overall very nice. The tests for the valuesprovider/controlplane controller needs to be adapted. They fail because the new property `nodeVolumeAttachLimit` is missing here: https://github.com/gardener/gardener-extension-provider-openstack/blob/master/pkg/controller/controlplane/valuesprovider_test.go#L247-L262 

We should also mention in the [operator docs](https://github.com/gardener/gardener-extension-provider-openstack/blob/master/docs/usage-as-operator.md#cloudprofileconfig) that this feature will only be available when using CSI.

mganter:
Added docs. Updated test.

dkistner (approved):

rfranzke (approved):

rfranzke:
@mganter can you check the failing CI? probably you need to run `make generate`

mganter:
> @mganter can you check the failing CI? probably you need to run `make generate`
> /status author-action

`make generate` results in no change of code. Removing `controller-registration.yaml` results in recreating it exactly with no diff. I rebased it to the latest rev. 

Some system information:
I'm running Manjaro Kernel: Linux 5.8.11-1-MANJARO
ldd (GNU libc) 2.32

Do you know if there are any issues with that?

rfranzke:
Now there is a conflict.

> Do you know if there are any issues with that?

Hm, good question, can you run `make install-requirements` and try again?

mganter:
The conflict is resolved now. (Didn't have the changes from 1 hour ago)
And its re-generated.

rfranzke (approved):

rfranzke:
Thanks @mganter!

mganter:
Thank you very much  @rfranzke @dkistner

PR #169, opened 2020-10-27
Reuse existing network

### LINKED ISSUE #99: Shoots in existing networks
/kind enhancement
/priority normal
/platform openstack

Like for other cloud providers we want to enable OpenStack Shoot deployments into existing networks. As we do it on the other infrastructures we will manage subnet as smallest unit in the network and it need to be disjunct from the other subnets in the network.

A example configuration could look like this.
```yaml
kind: InfrastructureConfig
networks:
  id: <id-of-my-network>
  workers: <cidr-of-subnet-in-my-network>
```

MrBatschner:

/area networking
/kind enhancement
/priority normal
/platform openstack

This PR enhances the gardener extension provider for OpenStack to be able to deploy shoots into existing networks.

Fixes #99 

This PR adds the network name field to the infratructure API and enables the terraformer to skip the network creation part. All MCM related activities will follow in a separate PR which will be linked to this one. 

**Release note**:

```improvement user
An existing Neutron network can be selected in a shoot's infrastructure configuration in the .spec.provider.infrastructureConfig.network.name field to have the shoot deployed into that network.
```

### DIFF
--- a/pkg/apis/openstack/types_infrastructure.go
+++ b/pkg/apis/openstack/types_infrastructure.go
@@ -40,6 +40,8 @@ type Networks struct {
 	Worker string
 	// Workers is a CIDRs of a worker subnet (private) to create (used for the VMs).
 	Workers string
+	// ID is the name of an existing network that should be reused
+	ID *string
 }
 
 // Router indicates whether to use an existing router or create a new one.
--- a/pkg/apis/openstack/v1alpha1/types_infrastructure.go
+++ b/pkg/apis/openstack/v1alpha1/types_infrastructure.go
@@ -43,6 +43,9 @@ type Networks struct {
 	Worker string `json:"worker"`
 	// Workers is a CIDRs of a worker subnet (private) to create (used for the VMs).
 	Workers string `json:"workers"`
+	// ID is the OpenStack UUID of an existing network that should be reused
+	// +optional
+	ID *string `json:"id,omitempty"`
 }
 
 // Router indicates whether to use an existing router or create a new one.
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -124,6 +124,11 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 		return err
 	}
 
+	subnet, err := helper.FindSubnetByPurpose(infrastructureStatus.Networks.Subnets, api.PurposeNodes)
+	if err != nil {
+		return err
+	}
+
 	for _, pool := range w.worker.Spec.Pools {
 		zoneLen := int32(len(pool.Zones))
 
@@ -155,6 +160,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 				"machineType":      pool.MachineType,
 				"keyName":          infrastructureStatus.Node.KeyName,
 				"networkID":        infrastructureStatus.Networks.ID,
+				"subnetID":         subnet.ID,
 				"podNetworkCidr":   extensionscontroller.GetPodNetwork(w.cluster),
 				"securityGroups":   []string{nodesSecurityGroup.Name},
 				"tags": map[string]string{
--- a/pkg/internal/infrastructure/terraform.go
+++ b/pkg/internal/infrastructure/terraform.go
@@ -69,8 +69,9 @@ func ComputeTerraformerChartValues(
 	cluster *controller.Cluster,
 ) (map[string]interface{}, error) {
 	var (
-		createRouter = true
-		routerConfig = map[string]interface{}{
+		createRouter  = true
+		createNetwork = true
+		routerConfig  = map[string]interface{}{
 			"id": DefaultRouterID,
 		}
 		outputKeysConfig = map[string]interface{}{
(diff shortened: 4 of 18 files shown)


MrBatschner:
Drafted this PR as changes to MCM need to be submitted first...

pkg/apis/openstack/v1alpha1/types_infrastructure.go
```diff
@@ -43,6 +43,9 @@ type Networks struct {
 	Worker string `json:"worker"`
 	// Workers is a CIDRs of a worker subnet (private) to create (used for the VMs).
 	Workers string `json:"workers"`
+	// Name is the name of an existing network that should be reused
```
dkistner, line 46:
```suggestion
	// Name is the name of an existing private network that should be reused.
```

pkg/apis/openstack/v1alpha1/types_infrastructure.go
```diff
@@ -43,6 +43,9 @@ type Networks struct {
 	Worker string `json:"worker"`
 	// Workers is a CIDRs of a worker subnet (private) to create (used for the VMs).
 	Workers string `json:"workers"`
```
dkistner, line 45: Hmm I guess in consequence of adding a choice (existing network or new network created by Gardener) we need to make this as well a `*string` and `+optional`.

I guess we need then also to validate that either `.workers` or `.name` is set. One of those is required.

pkg/apis/openstack/types_infrastructure.go
```diff
@@ -40,6 +40,8 @@ type Networks struct {
 	Worker string
 	// Workers is a CIDRs of a worker subnet (private) to create (used for the VMs).
 	Workers string
```
dkistner, line 42: See comments in `pkg/apis/openstack/v1alpha1/types_infrastructure.go`

pkg/internal/infrastructure/terraform.go
```diff
@@ -111,6 +119,7 @@ func ComputeTerraformerChartValues(
 	if workersCIDR == "" {
 		workersCIDR = config.Networks.Worker
 	}
+	networkConfig["workers"] = workersCIDR
```
dkistner, line 122: Do we need this key/value only if `config.Networks.Name == nil`?

pkg/controller/worker/machines.go
```diff
@@ -171,6 +176,7 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 				"machineType":      pool.MachineType,
 				"keyName":          infrastructureStatus.Node.KeyName,
 				"networkID":        infrastructureStatus.Networks.ID,
+				"subnetID":         subnet.ID,
```
dkistner, line 179: Do we need to pass the `subnetID` always? I mean we need it indeed for Shoots in existing networks. But what about all other Shoots with are currently deployed into Gardener managed networks. Does it have an impact on the existing Shoots when we now always add the `subnetID` to their machine classes?

dkistner (changes requested):
Meanwhile a mcm release which contain this PR https://github.com/gardener/machine-controller-manager/pull/545 is integrated. We can therefore remove the draft state of this PR.

We should add some documentation for this new deployment option in `docs/usage-as-end-user.md`.

I just a had a first look and wanted to play around with it a bit more.
Just added so far a few first comments.

pkg/apis/openstack/v1alpha1/types_infrastructure.go
MrBatschner: What we are specifying is the name of an existing Neutron network. The CIDR will then be used to create a new subnet in the network we previously specified so `Workers` is still a mandatory field IMHO.

pkg/apis/openstack/types_infrastructure.go
MrBatschner: see reply to said comment

pkg/internal/infrastructure/terraform.go
MrBatschner: As we do not consider the `Workers` CIDR to be an optional field, that assignment would always be needed.

pkg/apis/openstack/v1alpha1/types_infrastructure.go
dkistner: Ahh sorry. I mixed things. You are right. I thought `.workers` was the cidr for the network, but on OpenStack networks does not have an own cidr or at least we don't specify it.

pkg/internal/infrastructure/terraform.go
dkistner: Yes right, as pointed out in the comment above.

MrBatschner:
Still keeping this PR in draft as it needs some more tweaking. Two problems we discovered during extended testing:

- Specifying a network to reuse _by name_ is a fairly bad idea as OpenStack allows to have multiple resources with the same name. If there were two networks with the same name (but different UUIDs), the infrastructure reconciliation would fail at a very early stage with no chance of recovery. We have to change the code to deal so that the Neutron network needs to be given _by its unique ID_.
- A more challenging problem we discovered:
  - one manually created OpenStack _network_, with one _subnet_ in it (a.k.a. _subnet one_)
  - this _subnet_ is empty with no VMs and no router to a FIP network.
  - new shoot gets created in this _network_
  - terraformer creates a new _subnet_ in this _network_ with a router to a FIP network (a.k.a. _subnet two_)
  - new VMs get deployed to _subnet two_
  - VMs **cannot** access the network and thus not join the cluster
  - manually creating a router from _subnet one_ to a FIP network -> VMs in _subnet two_ can access the network
  - manually deleting the router from _subnet one_ -> VMs on _subnet two_ can no longer access the network
  - manually deleting (empty) _subnet one_ -> VMs on _subnet two_ can access the network

Not sure if this is expected behaviour but if it is, we clearly need to document that one must not have _subnets_ without a router to a FIP network in the Neutron _network_ that Shoots should be deployed to.

kon-angelo:
Closed in favor of the more recent: https://github.com/gardener/gardener-extension-provider-openstack/pull/317

PR #170, opened 2020-10-28
add servergroup for workerpools

### LINKED ISSUE #163: ServerGroup per Workerpool to define Node Affinity/Anti-Affinity

/area open-source
/kind enhancement
/priority normal
/platform openstack

We want to allow to assign each workerpool of an Openstack Shoot to assign a Affinity or Anti-Affinity Policy. If a policy is assigned to the workerpool then we need to create a OpenStack Servergroup for the workerpool which ensures the policy for the machines of the workerpool.

**Why is this needed**:
We need to ensure that not all machines of a workerpool get scheduled to one workerpool (node anti-affinity). We wanted to make let the user define which Openstack ServerGroup policy should be assigned to the workerpool. Gardener operators will define in the Cloudprofile which policies are valid.

kon-angelo:

/area control-plane
/kind enhancement
/priority normal
/platform openstack

Allows to create worker pool workers in server groups. Server-groups can have a defined affinity that will protect the workers from being scheduled in the same hypervisor, thus improving HA setups.

**Which issue(s) this PR fixes**:1
Fixes #163 

As part of the reconciliation loop we add the capability to create server groups in the `DeployMachineDependencies` function. Background and discussion topics for some of the design choices follow:

+ We are mostly interested in introducing `soft-anti-affinity` and `anti-affinity`policies.
+ Openstack APIs e.g. Nova aside from the major version also have a `microversion` which changes the API behavior. The client uses the lowest supported microversion by default, but `soft-*` affinity variants may be introduced in a later microversion  like in ConvergedCloud's case. Our implementation parses and uses the highest supported MV, but because this may not be supported in all OS environments, we opted to allow admins to set allowed policies in `CloudProfile`. 
+ Servergroups by default have no update operations. Any requested change e.g to the policy would mean to destroy and recreate the server group (and make a rolling update of nodes). 
+ Based on the previous point we opted to make the following implementation for the introduction of the feature. The validation of the servergroups will check to see if the `WorkerPoolProviderConfig` is immutable, disabling any update to the servergroup itself. 
+ Internally though, the reconciliation loop supports such operations. If the spec is changed a new server group will be created and new `machineclasses` will be created with a new hash. After the rolling node update, the old servergroup will eventually be disposed of. The update operation is only disabled by validation at the moment.
+ Servergroups is a project-wide object, but openstack doesn't support labels in objects, thus making it difficult for us to recognize our resources vs user-created-resources. In order to avoid conflicts in naming by multiple shoots in a project we use the resource's `Name` to create a labeling scheme. Servergroups are thus named as `clusterName-workerPoolName-<random-10char>`. The maximum allowed names are 255 chars so there is plenty of room to increase the random component. But as part of the naming scheme any server group in the OS project that has a prefix as `clusterName-workerPoolName` is thought to be managed by gardener and subject to deletion. 

Merge is pending update in MCM to include the servergroup scheduler [changes](https://github.com/gardener/machine-controller-manager/commit/0094e6331f8ddc1e224d6ca2f55802f7f90094bc)

**Release note**:

```improvement operator
Adds an additional option for the worker pools to specify a server group policy. If this option is set, a new server group with the defined policy will be created and nodes managed by the worker pool will become members. Allowed policy values can be defined in the provider's `CloudProfile`.
```


dkistner:
Awesome. Thanks. Will review it soon :)

docs/usage-as-end-user.md
```diff
@@ -81,6 +81,22 @@ The `cloudControllerManager.featureGates` contains a map of explicitly enabled o
 For production usage it's not recommend to use this field at all as you can enable alpha features or disable beta/stable features, potentially impacting the cluster stability.
 If you don't want to configure anything for the `cloudControllerManager` simply omit the key in the YAML specification.
 
+## `WorkerPoolProviderConfig`
```
rfranzke, line 84:
```suggestion
## `WorkerConfig`
```

pkg/apis/openstack/helper/scheme.go
```diff
@@ -70,3 +69,18 @@ func CloudProfileConfigFromCluster(cluster *controller.Cluster) (*api.CloudProfi
 	}
 	return cloudProfileConfig, nil
 }
+
+func WorkerPoolProviderConfigFromRawExtension(raw *runtime.RawExtension) (*api.WorkerPoolProviderConfig, error) {
```
rfranzke, line 73: doc string

pkg/apis/openstack/types_worker.go
```diff
@@ -43,3 +46,30 @@ type MachineImage struct {
...
+	PoolName string
+	// ID is the provider's generated ID for a server group
+	ID string
+	// Name is the name of the server group
+	Name string
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// WorkerPoolProviderConfig contains configuration data for a worker pool.
+type WorkerPoolProviderConfig struct {
```
rfranzke, line 62:
```suggestion
type WorkerConfig struct {
```

pkg/apis/openstack/v1alpha1/types_worker.go
```diff
@@ -32,6 +32,10 @@ type WorkerStatus struct {
 	// reconciliation is possible.
 	// +optional
 	MachineImages []MachineImage `json:"machineImages,omitempty"`
+
+	// +optional
+	// ServerGroupDependencies is a list of external server group dependencies.
```
rfranzke, line 37:
```suggestion
	// ServerGroupDependencies is a list of external server group dependencies.
	// +optional
```

pkg/apis/openstack/v1alpha1/types_worker.go
```diff
@@ -45,3 +49,30 @@ type MachineImage struct {
...
+	Name string `json:"name"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// WorkerPoolProviderConfig contains configuration data for a worker pool.
+type WorkerPoolProviderConfig struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// ServerGroupConfig contains configuration data for the worker pool's server group. If this object is present,
+	// Openstack extension will try to create a new server group for instances of this worker pool.
```
rfranzke, line 69:
```suggestion
	// OpenStack extension will try to create a new server group for instances of this worker pool.
	// +optional
```

pkg/apis/openstack/validation/shoot.go
```diff
@@ -60,9 +72,46 @@ func ValidateWorkersUpdate(oldWorkers, newWorkers []core.Worker, fldPath *field.
...
 					allErrs = append(allErrs, apivalidation.ValidateImmutableField(newWorker.Zones, oldWorker.Zones, fldPath.Index(i).Child("zones"))...)
 				}
-				break
+
+				if newWorker.ProviderConfig != nil && oldWorker.ProviderConfig != nil {
+					allErrs = append(allErrs, apivalidation.ValidateImmutableField(string(newWorker.ProviderConfig.Raw), string(oldWorker.ProviderConfig.Raw), fldPath.Index(i).Child("providerConfig"))...)
+				} else if newWorker.ProviderConfig != nil {
+					allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("providerConfig"), string(newWorker.ProviderConfig.Raw), apivalidation.FieldImmutableErrorMsg))
+				} else if oldWorker.ProviderConfig != nil {
+					allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("providerConfig"), nil, apivalidation.FieldImmutableErrorMsg))
```
rfranzke, line 81: The whole provider config is immutable? This makes it hard to extend it later. Probably it's better to just make the fields inside the provider config immutable.

pkg/controller/worker/helper.go
```diff
@@ -58,3 +64,44 @@ func (w *workerDelegate) updateWorkerProviderStatus(ctx context.Context, workerS
...
+	if pool.ProviderConfig == nil || pool.ProviderConfig.Raw == nil {
+		return nil, nil
+	}
+
+	pc, err := helper.WorkerPoolProviderConfigFromRawExtension(pool.ProviderConfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not decode WorkerPoolProviderConfig '%s/%s'", kutil.ObjectName(w.worker), pool.Name)
+	}
+
+	return pc, nil
```
rfranzke, line 78: maybe 
```suggestion
	return helper.WorkerPoolProviderConfigFromRawExtension(pool.ProviderConfig)
```

pkg/controller/worker/helper.go
```diff
@@ -58,3 +64,44 @@ func (w *workerDelegate) updateWorkerProviderStatus(ctx context.Context, workerS
...
+		} else {
+			err = statusUpdateErr
+		}
+	}
+
+	return err
+}
+
+// ClusterTechnicalName returns the technical name of the cluster this worker manages
+func (w *workerDelegate) ClusterTechnicalName() string {
```
rfranzke, line 95: Why do we need a function for this?

pkg/controller/worker/machines.go
```diff
@@ -231,3 +245,19 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
...
+	if serverGroupDependency != nil {
+		additionalHashData = append(additionalHashData, serverGroupDependency.Name, serverGroupDependency.ID)
+	}
+
+	// Generate the worker pool hash.
+	workerPoolHash, err := worker.WorkerPoolHash(pool, w.cluster, additionalHashData...)
+	if err != nil {
+		return "", err
+	}
+	return workerPoolHash, nil
```
rfranzke, line 262:
```suggestion
	return worker.WorkerPoolHash(pool, w.cluster, additionalHashData...)
```

pkg/controller/worker/server_groups.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	return result
+}
+
+type serverGroupDependencySet struct {
+	set map[string]api.ServerGroupDependency
+}
+
+func ServerGroupDependencySet(deps []api.ServerGroupDependency) serverGroupDependencySet {
```
rfranzke, line 57: doc string

pkg/openstack/client/client.go
```diff
@@ -0,0 +1,140 @@
...
+
+	os "github.com/gardener/gardener-extension-provider-openstack/pkg/openstack"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/utils/openstack/clientconfig"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NewOpenstackClientFromCredentials(credentials *os.Credentials) (Factory, error) {
```
rfranzke, line 30: doc string

pkg/openstack/client/client.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &OpenstackClientFactory{
+		providerClient: provider,
+	}, nil
+}
+
+func NewOpenstackClientFromSecretRef(ctx context.Context, c client.Client, secretRef corev1.SecretReference) (Factory, error) {
```
rfranzke, line 61: doc string

pkg/openstack/client/client.go
```diff
@@ -0,0 +1,140 @@
...
+	creds, err := os.GetCredentials(ctx, c, secretRef)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewOpenstackClientFromCredentials(creds)
+}
+
+func Region(region string) Option {
```
rfranzke, line 70: doc string
```suggestion
func WithRegion(region string) Option {
```

pkg/openstack/client/client.go
```diff
@@ -0,0 +1,140 @@
...
+	return func(opts gophercloud.EndpointOpts) gophercloud.EndpointOpts {
+		opts.Region = region
+		return opts
+	}
+}
+
+func (oc *OpenstackClientFactory) Storage(options ...Option) (Storage, error) {
```
rfranzke, line 77: doc string

pkg/openstack/client/client.go
```diff
@@ -0,0 +1,140 @@
...
+		client: storageClient,
+	}, nil
+}
+
+type serviceVersion struct {
+	Version struct {
+		Version    string `json:"version"`
+		Status     string `json:"status"`
+		MinVersion string `json:"min_version"`
+	} `json:"version"`
```
rfranzke, line 97: Why are the fields exported?
Why is this an inline struct leading to json paths like `version.version`?

pkg/openstack/client/client.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (oc *OpenstackClientFactory) Compute(options ...Option) (Compute, error) {
```
rfranzke, line 100: doc string

pkg/openstack/client/client.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil, err
+	}
+	client.Microversion = version.Version.Version
+
+	return &ComputeClient{
+		client: client,
+	}, nil
+}
+
+func IsNotFoundError(err error) bool {
```
rfranzke, line 127: doc string

pkg/openstack/client/compute.go
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package client
+
+import (
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/servergroups"
+)
+
+func (c *ComputeClient) CreateServerGroup(name, policy string) (*servergroups.ServerGroup, error) {
```
rfranzke, line 23: doc string for all exported functions

pkg/openstack/client/types.go
```diff
@@ -12,22 +12,48 @@
...
+type OpenstackClientFactory struct {
+	providerClient *gophercloud.ProviderClient
+}
+
+// StorageClient represents an Openstack storage client.
 type StorageClient struct {
 	client *gophercloud.ServiceClient
 }
 
+type ComputeClient struct {
```
rfranzke, line 35: doc strings again

pkg/controller/worker/helper.go
kon-angelo: Arguably unnecessary. 

Sometimes I like to use functions as a small wrapper of context, in this case conveying the concept that the names of the resources we create have a special convention and used for pseudo-labeling. Otherwise someone that reads the code for the first time could find it confusing why there is a reference to the `w.cluster.objectMeta.name` in a context that it is not directly related to the cluster resource.

pkg/openstack/client/client.go
kon-angelo: a) The fields are exported because the AFAIK the JSON decode only takes into account (thus filling) exported fields.
b) I was confused too but that's what openstack returns for the call. The struct is a 1-to-1 match. [Source](1), but the call we use does not return a list and the return json looks like: 
```json
{
  "version": {
      {
          "id": "v2.0",
          "links": [
              {
                  "href": "http://openstack.example.com/v2/",
                  "rel": "self"
              }
          ],
          "status": "SUPPORTED",
          "version": "",
          "min_version": "",
          "updated": "2011-01-21T11:33:21Z"
      }
}
```
As for inlining the struct, its used and will be used only in this place to read the response. My reasoning was to not introduce symbols with practically 0 reusability. 

The naming of the fields is to avoid `VersionOuter` variants and create more confusion.

[1]: https://docs.openstack.org/api-guide/compute/microversions.html

pkg/apis/openstack/validation/shoot.go
kon-angelo: There was no provider config before this change, thus making it immutable is semantically equivalent to making the fields immutable. I opted for this, so as to not introduce complexity which is not needed immediately. But yes, should we need to extend the config, then we would need to make the fields immutable and I can do that now.

rfranzke: Yes, let's do it now while we are on it anyways, this makes further contributions easier!

pkg/openstack/client/client.go
rfranzke: Interesting, thanks for clarifying!

pkg/controller/worker/helper.go
rfranzke: OK, good argument, thanks!

dkistner:
Can we move the PR to draft until the depending MCM change is integrated? https://github.com/gardener/machine-controller-manager/commit/0094e6331f8ddc1e224d6ca2f55802f7f90094bc

kon-angelo:
Main changes on 2nd revision: 
+ Rewrote update validation following @rfranzke suggestions. 
+ One additional validation rule has been added for shoot creation. Using "hard" affinity on workers with multiple availability zones is now forbidden since it results in impossible scheduling. (thanks to @dkistner for helping discover this)
+ cloudprofile validation
+ docstrings and export scoping improvements .

kon-angelo:
Since we have the new MCM release this PR is now open for merging. I just rebased on current master and resolved conflicts.

pkg/controller/worker/helper.go
```diff
@@ -58,3 +62,31 @@ func (w *workerDelegate) updateWorkerProviderStatus(ctx context.Context, workerS
...
+
+	return err
+}
+
+// ClusterTechnicalName returns the technical name of the cluster this worker belongs.
+func (w *workerDelegate) ClusterTechnicalName() string {
+	return w.cluster.ObjectMeta.Name
+}
+
+func (w *workerDelegate) findMatchingServerGroupPolicy(policy string, allowedPolicies []string) bool {
```
rfranzke, line 84: Does it make sense to move this into the API helper package? (actually, is this even called somewhere? I cannot find it 🔍)

pkg/controller/worker/machine_dependencies.go
```diff
@@ -1,25 +1,181 @@
...
+	}
+
+	if poolProviderConfig == nil || poolProviderConfig.ServerGroupConfig == nil || poolProviderConfig.ServerGroupConfig.Policy == "" {
+		return nil, nil
+	}
+
+	poolDep := set.getByPoolName(pool.Name)
+	if poolDep != nil {
+		serverGroup, err := computeClient.GetServerGroup(poolDep.ID)
+		if err != nil && !osclient.IsNotFoundError(err) {
```
rfranzke, line 69: What's the expected behaviour when a "not found" error is returned? At the moment, it's silently ignored, right? Just to double-check that one.

docs/usage-as-end-user.md
```diff
@@ -84,6 +84,22 @@ The `cloudControllerManager.featureGates` contains a map of explicitly enabled o
...
+apiVersion: openstack.provider.extensions.gardener.cloud/v1alpha1
+kind: WorkerConfig
+serverGroupConfig:
+  policy: soft-anti-affinity
+```
+
+Currently, the only configuration available is the `serverGroupConfig`. If this option is set, then a new server group will be created with the configured policy and all machines managed by this worker pool will be assigned as members of the group.
+
+Please note that the available options for `serverGroupConfig` are defined in the provider specific section of `CloudProfile`.
+
```
rfranzke, line 102: Can you also explain whether it's possible to add/change/remove this for existing clusters/worker pools?

pkg/controller/worker/machine_dependencies.go
```diff
@@ -1,25 +1,181 @@
...
+}
+
+// cleanupServerGroupDependencies handles deletion of excess server groups managed by the worker.
+// Cases handled:
+// a) worker is terminating and all server groups have to be deleted
+// b) worker pool is deleted
+// c) worker pool's server group configuration (e.g. policy) changed
+// d) worker pool no longer requires use of server groups
+func (w *workerDelegate) cleanupServerGroupDependencies(computeClient osclient.Compute, set serverGroupDependencySet) error {
+
```
rfranzke, line 120:
```suggestion
```

pkg/controller/worker/machine_dependencies_test.go
```diff
@@ -0,0 +1,469 @@
+package worker_test
```
rfranzke, line 1: license header missing

rfranzke (changes requested):
Very nice PR, great job!

pkg/controller/worker/helper.go
```diff
@@ -34,8 +34,12 @@ func (w *workerDelegate) decodeWorkerProviderStatus() (*api.WorkerStatus, error)
...
 	}
 
-	if _, _, err := w.Decoder().Decode(w.worker.Status.ProviderStatus.Raw, nil, workerStatus); err != nil {
-		return nil, errors.Wrapf(err, "could not decode WorkerStatus '%s'", kutil.ObjectName(w.worker))
+	marshalled, err := w.worker.Status.GetProviderStatus().MarshalJSON()
+	if err != nil {
+		return nil, err
+	}
+	if _, _, err := w.Decoder().Decode(marshalled, nil, workerStatus); err != nil {
+		return nil, errors.Wrapf(err, "could not decode WorkerStatus %q", kutil.ObjectName(w.worker))
```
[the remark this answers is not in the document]
kon-angelo: If you mean why using the `	marshalled, err := w.worker.Status.GetProviderStatus().MarshalJSON()` the straight answer is that it makes testing a bit easier. 

The `MarshalJson` function unifies RawExtensions's `Raw` and `Object` by returning the `Raw` is not empty and otherwise marshals the `Object` and returns that. That way I can skip the few lines of code in tests needed to write a struct/instance and marshal it into Raw. 

Normally on production only the Raw is populated, so the behaviour is the same. For all the other places tests or internal use, I think its great that I do not need to check where the data are (raw vs object) and always make sure I can access whatever is in a RawExtension without extra checks

pkg/controller/worker/helper.go
kon-angelo: Nice catch. That's an artifact from before I add the validation for the worker specs

pkg/controller/worker/machine_dependencies.go
kon-angelo: Yes, we silently ignore it. We will create a new server group and overwrite the old serverGroupID with the new in our worker's dependencies status. 

Later in the worker reconciliation we will garbage collect all serverGroups that match our "name-tag" `clusterName-workerName` but have no reference in our dependencies status.

pkg/controller/worker/helper.go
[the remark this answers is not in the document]
rfranzke: Interesting, thank you!

pkg/controller/worker/machine_dependencies.go
rfranzke: OK, thanks

docs/usage-as-end-user.md
```diff
@@ -96,9 +96,14 @@ serverGroupConfig:
   policy: soft-anti-affinity
 ```
 
-Currently, the only configuration available is the `serverGroupConfig`. If this option is set, then a new server group will be created with the configured policy and all machines managed by this worker pool will be assigned as members of the group.
-
-Please note that the available options for `serverGroupConfig` are defined in the provider specific section of `CloudProfile`.
+When you specify `serverGroupConfig` in your worker configuration, a new server group will be created with the configured policy and all machines managed by this worker will be assigned as members of the created server group.
```
rfranzke, line 99: As an end-user I would be interested if that happens in-place of whether my existing machines will be terminated/replaced by new machines (i.e., rolling update)?

docs/usage-as-end-user.md
```diff
@@ -96,9 +96,14 @@ serverGroupConfig:
...
+
+Please note the following restrictions when deploying workers with server groups:
++ The `serverGroupConfig` section is optional, but if it is included in the shoot spec, it must contain a valid policy value.
++ The available `policy` values that can be used, are defined in the provider specific section of `CloudProfile`, by your operator.
++ Certain policy values may induce further constraints. Using the `affinity` value is only allowed when the workers utilize a single zone.
++ The `serverGroupConfig` and `serverGroupConfig.policy` fields are immutable upon creation. Users are not allowed to change the policy value and neither add or remove a `serverGroupConfig` section from a worker after it has been created.
```
rfranzke, line 105: > from a worker

from a worker pool

docs/usage-as-end-user.md
```diff
@@ -96,9 +96,14 @@ serverGroupConfig:
...
+Users can on the other hand, add new workers managed by server groups to an existing shoot and migrate their workloads to the new workers.
```
rfranzke, line 106: OK, now reading this, the first sentence (with my comment about in-place) is clarified. Maybe this can be made more clear from the beginning, i.e., that all of this only applies for newly created worker pools while existing pools are immutable.

pkg/controller/worker/machine_dependencies.go
```diff
@@ -1,25 +1,180 @@
...
+	return w.updateMachineDependenciesStatus(ctx, workerStatus, serverGroupDepSet.extract(), err)
+}
+
+// cleanupServerGroupDependencies handles deletion of excess server groups managed by the worker.
+// Cases handled:
+// a) worker is terminating and all server groups have to be deleted
+// b) worker pool is deleted
+// c) worker pool's server group configuration (e.g. policy) changed
+// d) worker pool no longer requires use of server groups
```
dkistner, line 118: Is case d) actually possible? I though the servergroup + policy are immutable and can't be changed. I mean anyway the handling is the same.

pkg/controller/worker/machines.go
```diff
@@ -205,3 +219,15 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 
 	return nil
 }
+
+func (w *workerDelegate) generateWorkerPoolHash(pool extensionsv1alpha1.WorkerPool, serverGroupDependency *api.ServerGroupDependency) (string, error) {
+	var additionalHashData = []string{}
+
+	// Include the given worker pool dependencies into the hash.
+	if serverGroupDependency != nil {
+		additionalHashData = append(additionalHashData, serverGroupDependency.Name, serverGroupDependency.ID)
```
dkistner, line 228: Out of curiosity: Wouldn't it be sufficient to just add the servergroup dependency `id` to the hash? I mean this one should be unique.

pkg/openstack/client/mocks/client_mocks.go
```diff
@@ -0,0 +1,156 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: github.com/gardener/gardener-extension-provider-openstack/pkg/openstack/client (interfaces: Factory,Compute)
+
+// Package mocks is a generated GoMock package.
+package mocks
```
dkistner, line 5: I'm wondering if we should move all mock packages to `pkg/mock` with subdirectories for the `client` or any other  mock. `g/g` is doing it also this way. https://github.com/gardener/gardener/tree/master/pkg/mock

pkg/openstack/client/types.go
```diff
@@ -12,22 +12,51 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:generate mockgen -destination=mocks/client_mocks.go -package=mocks . Factory,Compute
```
dkistner, line 15: Same question as in `pkg/openstack/client/mocks/client_mocks.go`

pkg/openstack/client/mocks/client_mocks.go
kon-angelo: What i don't like about this:
a) if we need changes to the provider and need to use a new openstack service not mentioned here, we would need a g/g release.
b) I do not consider these interfaces as stable to move them into the central repo. In comparison the mcm-provider-openstack uses a slightly different structure. If we consolidate them we could think about it but its premature now

pkg/controller/worker/machine_dependencies.go
kon-angelo: Controller fully supports c) and d) cases although these are prohibited in the validation for now. To support these the internal implementation does need to do a rolling node update by generating new MachineClasses. But generally we can disable the immutability checks at any point in the future and the deletion will work.

pkg/openstack/client/mocks/client_mocks.go
dkistner: My intention was not to move the mocks of the openstack extension to the g/g repo. I wanted to suggest to move all the mocks of the openstack extension to a central place like `pkg/mocks` in the openstack extension repo. Then the structure would be similar like in the g/g repo. :)

kon-angelo: I misunderstood. This is possible, yes. We can open an issue and give it a try

kon-angelo:
+ made documentation more explicit about migration to worker pools with server groups
+ explicit validation when the `serverGroup` is missing or added and added one more unit test for this
+ comments addressed

@rfranzke maybe can you check the documentation one more time. I used the term `worker groups` instead of pools because that's how its referred in the UI. I hope its better now

dkistner (approved):
After a pair review with @kon-angelo 
/lgtm

dkistner (approved):

PR #172, opened 2020-11-03
Clean up StorageClass update workaround

ialidzhikov:

/area storage
/kind cleanup
/priority normal
/platform openstack

Clean up the workaround introduced with https://github.com/gardener/gardener-extension-provider-openstack/pull/102.
I believe we no longer need it as StorageClass should be already adapted by the temporary workaround logic. We also introduced https://github.com/gardener/gardener-extension-provider-openstack/pull/130 which should re-create the StorageClasses on invalid updates.

**Release note**:

```improvement operator
NONE
```


rfranzke (approved):
/lgtm

Yes, makes sense, thanks!

PR #174, opened 2020-11-12
Changing the default parameters handling and adding all properties to the test definition

kris94:
How to categorize this PR?
/kind enhancement
/priority normal
/platform openstack

What this PR does / why we need it:
This PR adds all properties to the "provider-openstack" test definition. In addition to this, as discussed in [PR:191,](https://github.com/gardener/gardener-extension-provider-aws/pull/191) the default parameters handling is changed.
This way one test definition will be used by the integration tests where shoots with different networks from the defaults must be created (e.g. for registering a new seed for the control plane migration) and it will preserve the default behaviour for other integration tests (e.g. for shoot creation) that relies on the default values.
Which issue(s) this PR fixes:
Fixes #

Special notes for your reviewer:

Release note:
```
NONE
```


rfranzke (approved):
/lgtm

PR #185, opened 2020-11-30
update golang image to eu.gcr.io mirror

kon-angelo:

/area dev-productivity
/kind technical-debt
/priority normal
/platform openstack

Updates the images to the eu.grc.io golang image mirror and circumvent dockerhub rate limits

**Release note**:

```improvement operator
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,12 @@
 ############# builder
-FROM golang:1.15.3 AS builder
+FROM eu.gcr.io/gardener-project/3rd/golang:1.15.5 AS builder
 
 WORKDIR /go/src/github.com/gardener/gardener-extension-provider-openstack
 COPY . .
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 ############# builder
-FROM golang:1.15.3 AS builder
+FROM eu.gcr.io/gardener-project/3rd/golang:1.15.5 AS builder
```
ialidzhikov, line 2: Can you change also the alpine image below to `eu.gcr.io/gardener-project/3rd/alpine:3.12.1`? Just to be sure that we won't hit rate limit issue with the alpine image too.

rfranzke (approved):
/lgtm

dkistner (approved):
Thank you :)

PR #204, opened 2021-01-07
Only validate infra/controlPlane config against CloudProfile on change

rfranzke:

/area usability
/kind enhancement
/priority normal
/platform openstack

Now we only validate the `.spec.provider.{infrastructure,controlPlane}Config` values against the constraints defined in the `CloudProfile` if the values in the shoot were changed during an update. This is to ensure that constraints can be removed from the `CloudProfile` without breaking updates to `Shoot` resources.

**Release note**:

```bugfix operator
The validator does now only validate the `.spec.provider.{infrastructure,controlPlane}Config` values of a `Shoot` against the constraints in the `CloudProfile` if the values were changed during a `Shoot` update.
```


dkistner (approved):
/lgtm

PR #205, opened 2021-01-11
Use cinder-csi-plugin@v1.20.0 for Kubernetes >= 1.20

ialidzhikov:
/area storage
/platform openstack

**Release note**:

```other user
provider-openstack is now using cinder-csi-plugin@v1.20 for Kubernetes >= 1.20 clusters.
```


rfranzke (approved):
/lgtm

PR #212, opened 2021-01-25
remove server group update validation

kon-angelo:

/area control-plane
/kind enhancement
/priority normal
/platform openstack

This PR removes the validation checks that prevented worker groups to change server group settings after their creation. This will trigger a rolling update (by changing the `MachineClass` hash)

Fixes #

**Release note**:

```doc user
Allow updating server group settings on existing worker groups.
```

### DIFF
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -86,7 +86,7 @@ If you don't want to configure anything for the `cloudControllerManager` simply 
 
 ## `WorkerConfig`
 
-Each worker group in a shoot may contain provider specific configurations and options. These are contained in the `providerConfig` section of a worker group.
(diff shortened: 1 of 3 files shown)


dkistner:
Could you add a `breaking user` release note with a hint for user that changing the server group policy will lead to rolling update of the worker pool.

Furthermore we should add documentation for that with a bold hint of the rolling update.

rfranzke:
Why is it a `breaking` change @dkistner ?

dkistner:
Sorry didn't mean `breaking` rather `doc` to make it prominent for users.

kon-angelo:
I will remove the draft, once documentation is also updated.

dkistner (approved):

PR #218, opened 2021-02-01
Add useragent header for api call issued by ccm and csi driver

### LINKED ISSUE #214: Add a "--user-agent" CLI parameter to occm/csi components

/area control-plane
/kind enhancement
/priority normal
/platform openstack

`--user-agent` parameter adds an additional prefix to a `User-Agent` request header

**Why is this needed**:

If we set `--user-agent %openstack_domain_name% --user-agent %openstack_project_name% --user-agent %gardener_cluster_name%`, the resulting User-Agent header will look like:

```
User-Agent: %domain% %project% %cluster_name% occm/version gophercloud/2.0.0
``` 

This information will help us to quickly identify the suspect through HTTP access logs, when we face an issue.

dkistner:

/area ops-productivity
/kind enhancement
/priority normal
/platform openstack

Add the Openstack domain name, the tenant/project name and the Shoot technical name (Project name + Shoot name) as http headers for the API calls which the the Openstack Cloud Controller Manager and CSI drive run against OpenStack APIs.

This should help to detect log messages which belong to certain Shoot faster on the Openstack provider side.

Fixes #214

I wanted to play around before I remove the draft state.

**Release note**:

```other operator
The name of the Openstack domain, tenant/project name and the technical name of the respective Shoot will be added as http headers for the API call which the Openstack Cloud Controller Manager and CSI driver run against Openstack APIs.
```

cc @kayrus, @kon-angelo @dguendisch

### DIFF
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
@@ -12,6 +12,7 @@ images:
 
 socketPath: /var/lib/csi/sockets/pluginproxy
 timeout: 3m
+userAgentHeaders: []
 
 resources:
   driver:
--- a/charts/internal/shoot-system-components/charts/csi-driver-node/values.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-node/values.yaml
@@ -8,6 +8,7 @@ images:
 
 socketPath: /csi/csi.sock
 vpaEnabled: false
+userAgentHeaders: []
 
 resources:
   driver:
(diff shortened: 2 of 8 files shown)


kayrus:
LGTM

charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
```diff
@@ -50,6 +50,9 @@ resources:
       cpu: 30m
       memory: 50Mi
 
+csiDriver:
+  useragentHeaders: []
```
ialidzhikov, line 54: I would say that there is no need for csiDriver as the other values are already on the root level.

```suggestion
userAgentHeaders: []
```

charts/internal/shoot-system-components/charts/csi-driver-node/values.yaml
```diff
@@ -9,6 +9,9 @@ images:
 socketPath: /csi/csi.sock
 vpaEnabled: false
 
+csiDriver:
+  useragentHeaders: []
```
ialidzhikov, line 13:
```suggestion
userAgentHeaders: []
```

dkistner:
@ialidzhikov thanks for feedback. I have incorporated it.

ialidzhikov (approved):
/lgtm

PR #222, opened 2021-02-03
Add max retries to repeat failed or throttled operations

### LINKED ISSUE #220: Update openstack terraform provider configuration

/area usability
/kind enhancement
/priority normal
/platform openstack

Current provider configuration should be adjusted:

https://github.com/gardener/gardener-extension-provider-openstack/blob/91107343d96686ff9e59098d19536c0e2af3985c/charts/internal/openstack-infra/templates/main.tf#L1-L9

Insecure flag must be removed:

```hcl
insecure    = true
```

And a `max_retries` should be set in order support the 429 response code with a `Retry-After` header.

**Why is this needed**:

* Production environment must not contain the `insecure=true` parameter.
* `max_retries` starting from Terrafrom OpenStack provider v1.36 will support the 429 response code and sleep accordingly. This is needed in order to decrease high amount of API calls, when a technical user password was changed and a user gets blocked by identity.

dkistner:

/area robustness
/kind enhancement
/priority normal
/platform openstack

Add max retries for api call to retry failed or throttled operations.

Fixes partly #220

Require `terraformer-openstack` with provider openstack `v.1.37.0` integrated. See https://github.com/gardener/terraformer/pull/68

**Release note**:

```NONE

```

/invite @kon-angelo 
cc @kayrus

### DIFF
--- a/charts/internal/openstack-infra/templates/main.tf
+++ b/charts/internal/openstack-infra/templates/main.tf
@@ -6,6 +6,7 @@ provider "openstack" {
   user_name   = var.USER_NAME
   password    = var.PASSWORD
   insecure    = true
+  max_retries = "{{ required "openstack.maxApiCallRetries is required" .Values.openstack.maxApiCallRetries }}"
 }
 
 //=====================================================================
--- a/charts/internal/openstack-infra/values.yaml
+++ b/charts/internal/openstack-infra/values.yaml
@@ -4,6 +4,7 @@ openstack:
   tenantName: kubernetes
   region: eu-de-1
   floatingPoolName: my-pool
+  maxApiCallRetries: 10
 
 create:
   router: true


dkistner:
@kayrus How many retries would you recommend as default?

charts/internal/openstack-infra/templates/main.tf
```diff
@@ -5,7 +5,7 @@ provider "openstack" {
   region      = "{{ required "openstack.region is required" .Values.openstack.region }}"
   user_name   = var.USER_NAME
   password    = var.PASSWORD
-  insecure    = true
```
rfranzke, line 8: Removing this probably means to provide a way to pass the CA certificate for non-globally trusted server certs, or?

kayrus:
> How many retries would you recommend as default?

I think 10 is fine for now. In future this option should be used for OCCM/OS CSI drivers as well.

charts/internal/openstack-infra/templates/main.tf
kayrus: yes

dkistner: Hmm yes this was maybe to rushed and will not work in all openstack environments without providing the ability to pass the ca cert.

dkistner: Let's add `insecure=true` again and remove it with a subsequent PR where we also provide the ability to pass a proper CA cert to do the TLS. WDYT?

rfranzke: Agreed, makes sense. Otherwise, it will break in environments not under our control.

dkistner: done

dkistner:
As the new terraformer is now integrated I will remove the draft state.

kon-angelo (approved):
/lgtm

PR #228, opened 2021-02-17
Vendor gardener/gardener@v1.16.3

ialidzhikov:
/kind bug

**Release note**:

```other operator
An issue causing `github.com/gardener/gardener/pkg/utils/imagevector.FindImages` to not give a higher score on exact matched targetVersion or runtimeVersion is now fixed.
```

```other operator
An issue causing Shoots to be marked as `Failed` (and no longer retried) on transient `not found` error is now fixed.
```


dkistner (approved):
/lgtm

rfranzke (approved):

PR #239, opened 2021-03-10
Add controlplane label to csi-driver-controller

ialidzhikov:
`csi-driver-controller` Deployment is now annotated with the controlplane labels to allow the `dependency-watchdog` to restart the csi-driver-controller when the apiserver recovers.
Ref https://github.com/gardener/gardener/blob/v1.18.0/docs/usage/seed_bootstrapping.md#endpoint

**Release note**:

```other operator
NONE
```


dkistner (approved):
/lgtm

But wait for @kon-angelo to confirm if we can merge (he's validating a new release currently)

kon-angelo:
/lgtm

I don't think it will make any difference about the current validation. I can redeploy on top of the current one.

kon-angelo (approved):

PR #240, opened 2021-03-12
Update default storage provisioner to out-of-tree

morremeyer:
/area storage
/kind enhancement
/priority normal
/platform openstack

**What this PR does / why we need it**: Updates the default storage class to the out-of-tree `cinder.csi.openstack.org`

**Which issue(s) this PR fixes**: N/A

I think this should also update [example/controller-registration.yaml](example/controller-registration.yaml)s `spec.deployment.providerConfig.chart` value, but I’m not sure how the base64-encoded gzipped data in there is generated.

I’d like some input on that.

I also checked the docs for needed changes but the areas where storage is described can stay as they are in my opinion.

**Release note**:

```breaking operator
Update default provisioner for gardener.cloud-fast storage class to cinder.csi.openstack.org
```


gardener-robot-ci-2:
Thank you @morremeyer for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

kon-angelo:
Hello @morremeyer and thank you for the contribution. 

The `example/controller-registration.yaml` can be updated by running `make generate`.

morremeyer:
@kon-angelo Thanks! Unfortunately, `make generate` failed for me, so I just ran the needed command `./vendor/github.com/gardener/gardener/hack/generate.sh ./charts/...`.

I’ll open an issue for the fail.

charts/gardener-extension-provider-openstack/values.yaml
```diff
@@ -66,7 +66,7 @@ config:
     storage:
       className: gardener.cloud-fast
       capacity: 25Gi
-      provisioner: kubernetes.io/cinder
+      provisioner: cinder.csi.openstack.org
```
ialidzhikov, line 69: As far as I see this StorageClass is used for `etcd-main` PVs on the Seed. So the surprising thing for me was that the out-of-tree volume provisioners are compatible with the in-tree ones and for example can provision a PV that use a StorageClass with the old in-tree provisioner (I assume this is coming from the csi-translation magic). Which means that the out-of-tree driver can handle both `cinder.csi.openstack.org` and `kubernetes.io/cinder`.
So if do this change, I think it will break Seeds that don't support the `cinder.csi.openstack.org` provisioner. Also the `provisioner` field of StorageClass is immutable. So I assume the update will fail with failure.

So bottom line: we still would like to preserve compatibility with Seeds that only support `kubernetes.io/cinder`.

/hold

@rfranzke , any thoughts?

rfranzke: +1 @ialidzhikov, I guess it's not needed to change this (it's anyways just an example value that can be overridden).

ialidzhikov:
@morremeyer , I will close this PR because of https://github.com/gardener/gardener-extension-provider-openstack/pull/240#discussion_r617449508. Let us know if you have comments/objections. Thank you for your understanding!

PR #241, opened 2021-03-16
Use Terraform variables for credentials

rfranzke:

/area usability
/kind enhancement
/priority normal
/platform openstack

Instead of templating the credentials (`domainName`, `tenantName`) into the Terraform script we are now using container environment variables with references to the `cloudprovider` secret (similar to how it's done for `username` and `password`).
The Terraform script is only updated on creation/reconciliation, but if an `Infrastructure` is already marked for deletion then it's never updated again. This is a problem if the `domainName` or `tenantName` changes.

/cc @dguendisch @timuthy 

**Release note**:

```other operator
NONE
```


dguendisch:
Thank you!

dkistner (approved):
/lgtm

thank you.

kon-angelo (approved):
/lgtm

PR #245, opened 2021-03-24
fix failing servergroup creation on Nova 2.64+

kon-angelo:

/area control-plane
/kind bug
/priority blocker
/platform openstack

This PR fixes an issue when to trying to create servergroups with nova API `>= 2.64`. The request no longer accepts the `Policies` field on those versions. The current client picked the highest supported version of the Nova API and if that was greater than `2.64` the creation of servergroups would fail.

Instead, do not set the microversion of the client.  In case `soft-*` servergroup policies are requested, pin the client to `2.15`, since its the minimum version that supports these values.

Fixes #

**Release note**:

```bugfix operator
Fix a bug in servergroup creation when the Nova API is `> 2.63`
```


dkistner (approved):
/lgtm

PR #256, opened 2021-04-08
Update kubernetes-csi/livenessprobe

ialidzhikov:
/kind bug
/platform openstack

We observe a huge memory leak with the `csi-liveness-probe` sidecar. Its memory usage is increasing across the time.

Screenshot 2021-04-08 at 14 17 28

From the screenshot you can see that the memory for the `csi-liveness-probe` container goes above 1Gi.
There is upstream fix https://github.com/kubernetes-csi/livenessprobe/pull/94 for a memory leak that is available in `k8s.gcr.io/sig-storage/livenessprobe@v2.2.0`.

**Release note**:

```bugfix user
The following image is updated (see [CHANGELOG](https://github.com/kubernetes-csi/livenessprobe/blob/v2.2.0/CHANGELOG/CHANGELOG-2.2.md) for more details):
- k8s.gcr.io/sig-storage/livenessprobe: v2.1.0 -> v2.2.0
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -74,4 +74,4 @@ images:
 - name: csi-liveness-probe
   sourceRepository: github.com/kubernetes-csi/livenessprobe
   repository: k8s.gcr.io/sig-storage/livenessprobe
-  tag: "v2.1.0"
+  tag: "v2.2.0"


timebertt (approved):
/lgtm

PR #259, opened 2021-04-09
Specify container resources for CSI sidecars

ialidzhikov:
/kind enhancement
/platform openstack

Currently the liveness-probe and csi-node-driver-registrar sidecars on the csi-driver-node Pod, the liveness-probe sidecar on  the csi-driver-controller Pod do not define resource requests and limits.
Without any resource limits, currently we don't prevent any abnormal usage of resources by these sidecars. For example we recently observed a huge memory leak (see more details in #256) in the liveness-probe sidecar - because of the memory leak the sidecar container was using more than 1Gi when the normal (without the memory leak) memory usage of the container is ~20Mi. With specifying resource requests and limits, we can prevent such cases. 

**Release note**:

```other operator
The few CSI sidecar containers that didn't specify any resource requests and limits do now specify appropriate requests and limits. 
```

### DIFF
--- a/charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
+++ b/charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml
@@ -50,6 +50,13 @@ resources:
     limits:
       cpu: 30m
       memory: 50Mi
+  livenessProbe:
+    requests:
+      cpu: 10m
+      memory: 32Mi
+    limits:
+      cpu: 30m
+      memory: 50Mi
 
 csiSnapshotController:
   replicas: 1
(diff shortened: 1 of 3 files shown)


vpnachev (approved):
/lgtm

dkistner (approved):
/lgtm

kon-angelo (approved):
/lgtm

ialidzhikov:
> ```yaml
> resources:
> ```

Nice suggestion, but I don't want to adapt only 1 template file to use this `nindent` approach and to leave all other template files to use the `indent` approach. I don't want to adapt all templates under `chart/` neither as it is out of scope of this PR. For consistency reasons I will leave as it is.

PR #262, opened 2021-04-12
Align provider name with naming convension

prashanth26:

/area dev-productivity
/kind cleanup
/platform openstack

The provider name should match with the convension used by [machine controller openstack here](https://github.com/gardener/machine-controller-manager-provider-openstack/blob/master/pkg/driver/utils.go#L23). Specially with [this change](https://github.com/gardener/machine-controller-manager-provider-openstack/pull/19) coming in. 

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/machineclass/templates/machineclass.yaml
+++ b/charts/internal/machineclass/templates/machineclass.yaml
@@ -22,7 +22,7 @@ metadata:
   labels:
   {{ toYaml $machineClass.labels | indent 4 }}
 {{- end }}
-provider: "OpenStack"
+provider: "openstack"
 secretRef:
   name: {{ $machineClass.name }}
   namespace: {{ $.Release.Namespace }}


kon-angelo:
@prashanth26 I am a bit conflicted because in other providers e.g. [azure](https://github.com/gardener/gardener-extension-provider-azure/pull/257/files) , [aws](https://github.com/gardener/gardener-extension-provider-aws/blob/32dcc7ef152998192c9219bc85e5d693f0f5c99b/charts/internal/machineclass/templates/machineclass.yaml#L46), gcp we use camelcase. These were my point of reference too for the initial naming.

I would suggest we update the respective openstack-mcm PR to match that (plus it requires only one change to be commited). My argument is solely about consistency between the providers. 

If you feel a strong preference towards the lowercase we can go ahead and merge it though.

prashanth26:
> @prashanth26 I am a bit conflicted because in other providers e.g. [azure](https://github.com/gardener/gardener-extension-provider-azure/pull/257/files) , [aws](https://github.com/gardener/gardener-extension-provider-aws/blob/32dcc7ef152998192c9219bc85e5d693f0f5c99b/charts/internal/machineclass/templates/machineclass.yaml#L46), gcp we use camelcase. These were my point of reference too for the initial naming.
> 
> I would suggest we update the respective openstack-mcm PR to match that (plus it requires only one change to be commited). My argument is solely about consistency between the providers.
> 
> If you feel a strong preference towards the lowercase we can go ahead and merge it though.

💯 . Let's change the providerName to match with this -https://github.com/gardener/gardener-extension-provider-openstack/pull/262/files#diff-9b4849c5f886d8c9282bd3836aecbc442f1dc76bf89e3f7d91183b9758e0e646L25. 
cc: @AxiomSamarth

PR #286, opened 2021-05-21
Vendor gardener/gardener@v1.23.1

ialidzhikov:
**Release note**:

```other operator
NONE
```


example/controller-registration.yaml
```diff
@@ -3,13 +3,12 @@ apiVersion: core.gardener.cloud/v1beta1
...
-    values:
-      image:
-        tag: v1.20.0-dev
+type: helm
```
vpnachev, line 6: I guess needs cherry-pick ?

vpnachev (approved):
/lgtm

example/controller-registration.yaml
vpnachev: No, it does not. The last release is not using ControllerDeployment resource.

PR #294, opened 2021-06-10
Enhance validation of cloud provider secrets

stoyanr:

/area usability
/area ops-productivity
/kind enhancement
/platform openstack

Enhances the validation of Openstack cloud provider secrets:
* `domainName`, `tenantName`, and `userName` must not contain leading or trailing whitespace.
* `tenantName` must not be longer than 64 characters, see https://docs.openstack.org/api-ref/identity/v3/?expanded=show-project-details-detail.
* `password` must not contain leading or trailing new lines, as they are known to cause issues. Other whitespace characters such as spaces are intentionally not checked for, since there I couldn't find any documentation indicating that they would not be valid.
* `authURL` must be a valid URL if present. 

Fixes #

`authURL` is validated with `url.Parse` which is fairly lenient. I believe this is appropriate since there are no documented restrictions for this field besides being a "URL".

**Release note**:

```other operator
Validation of Openstack cloud provider secrets is enhanced to reject `domainName`, `tenantName`, and `userName` that contain leading or trailing whitespace, `tenantName` that is longer than 64 characters, `password` that contain leading or trailing new lines, and `authURL` that is not a valid URL.
```

### DIFF
--- a/pkg/apis/openstack/validation/secrets.go
+++ b/pkg/apis/openstack/validation/secrets.go
@@ -63,7 +63,7 @@ func ValidateCloudProviderSecret(secret *corev1.Secret) error {
 	// authURL must be a valid URL if present
 	if credentials.AuthURL != "" {
 		if _, err := url.Parse(credentials.AuthURL); err != nil {
-			return fmt.Errorf("field %q in secret %s must be a valid URL if present", openstack.AuthURL, secretKey)
+			return fmt.Errorf("field %q in secret %s must be a valid URL when present: %v", openstack.AuthURL, secretKey, err)
 		}
 	}
 
--- a/pkg/apis/openstack/validation/secrets.go
+++ b/pkg/apis/openstack/validation/secrets.go
@@ -15,12 +15,57 @@
 package validation
 
 import (
+	"fmt"
+	"net/url"
+	"strings"
+
 	"github.com/gardener/gardener-extension-provider-openstack/pkg/openstack"
+
 	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	tenantNameMaxLen = 64
 )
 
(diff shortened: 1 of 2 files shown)


pkg/apis/openstack/validation/secrets.go
```diff
@@ -15,12 +15,57 @@
...
+	// Other whitespace characters such as spaces are intentionally not checked for,
+	// since there is no documentation indicating that they would not be valid
+	if strings.Trim(credentials.Password, "\n\r") != credentials.Password {
+		return fmt.Errorf("field %q in secret %s must not contain leading or traling new lines", openstack.Password, secretKey)
+	}
+
+	// authURL must be a valid URL if present
+	if credentials.AuthURL != "" {
+		if _, err := url.Parse(credentials.AuthURL); err != nil {
+			return fmt.Errorf("field %q in secret %s must be a valid URL if present", openstack.AuthURL, secretKey)
```
vpnachev, line 66: I think it will be helpful to include the real error  in the message here. It can fail due to multiple reasons and it will be nice to know which one it was.
```suggestion
			return fmt.Errorf("field %q in secret %s must be a valid URL when present: %v", openstack.AuthURL, secretKey, err)
```

stoyanr: Good idea, thanks!

vpnachev (approved):
/lgtm
/needs second-opinion
from @gardener/gardener-extension-provider-openstack-maintainers

rfranzke (approved):
/lgtm
/squash
/invite @kon-angelo

kon-angelo (approved):
Sorry for the delay. Thank you @stoyanr 

/lgtm

PR #306, opened 2021-07-13
fix server group removal from worker pool

kon-angelo:

/area control-plane
/kind bug
/platform openstack

+ Fixes an issue with servergroups. Prior to the fix, when the servergroups are deleted from a workerpool it would require 2 reconcilliation loops for the change in the shoot's spec to create proper machineclasses.

Minor:
+ change a `map[string]struct{}` used for string lookup to `sets.String`
+ improve the servergroup reverse name lookup by taking into account a trailing `-`.  For example shoots like `shoot_name` and `shoot_name_2` will no longer collide during the reverse lookup. 

Fixes #

**Release note**:

```bugfix operator
Fixes an issue where removing server groups from a worker pool would not produce correct `machineclasses`. Prior to the fix, two shoot reconciliations would be necessary to reach the desired state.
```

### DIFF
--- a/pkg/controller/worker/machine_dependencies.go
+++ b/pkg/controller/worker/machine_dependencies.go
@@ -20,10 +20,12 @@ import (
 	"context"
 	"fmt"
 
+	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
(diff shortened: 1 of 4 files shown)


dkistner (approved):
/lgtm

PR #310, opened 2021-07-15
Improve LoadBalancerClass validation

dkistner:
/kind enhancement
/platform openstack

Allow to change the name and purpose of load balancer classes in `.controlPlaneConfig.loadBalancerClasses[]`. The load balancer classes configuration need still to be semantically equal with the load balancer classes from the CloudProfile.

Add validation to check if there are no load balancer classes with same name and if there are multiple default or private load balancer classes.

**Release note**:

```other user
It is now allowed to change the name and purpose of load balancer classes in `.controlPlaneConfig.loadBalancerClasses[]`. The load balancer classes configuration need still to be semantically equal with the load balancer classes from the CloudProfile.
```

/invite @kon-angelo @MartinWeindel 
cc @RaphaelVogel

### DIFF
--- a/pkg/apis/openstack/validation/cloudprofile.go
+++ b/pkg/apis/openstack/validation/cloudprofile.go
@@ -66,10 +66,7 @@ func ValidateCloudProfileConfig(cloudProfile *api.CloudProfileConfig, fldPath *f
 	}
 
 	for i, pool := range cloudProfile.Constraints.FloatingPools {
-		lbClassPath := floatingPoolPath.Index(i).Child("loadBalancerClasses")
-		for j, class := range pool.LoadBalancerClasses {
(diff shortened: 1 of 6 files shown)


pkg/apis/openstack/validation/cloudprofile.go
```diff
@@ -192,3 +189,47 @@ func ValidateLoadBalancerClasses(lbClass api.LoadBalancerClass, fldPath *field.P
 
 	return allErrs
 }
+
```
kon-angelo, line 192: Nit: Comment missing for exported function

dkistner: done. Thanks :)

kon-angelo (approved):
/lgtm

PR #312, opened 2021-07-15
Add DNSRecord controller

stoyanr:

/area control-plane
/kind api-change
/kind enhancement
/platform openstack

* Adds a Openstack-specific implementation of a DNSRecord controller, see https://github.com/gardener/gardener/pull/4231.
* Vendors g/g v1.27.0-dev and k8s v0.21.2 (in a separate commit to make reviews easier)

Fixes #

Note the following important details:
* The provider type is `openstack-designate` (like in `DNSProvider`), not `openstack` (like in other controllers) for reasons of compatibility and ease of migration.

See also https://github.com/gardener/gardener-extension-provider-aws/pull/362 and https://github.com/gardener/gardener-extension-provider-aws/pull/370.

**Release note**:

```other operator
NONE
```


pkg/openstack/credentials.go
```diff
@@ -41,33 +42,43 @@ type Credentials struct {
...
-	password := getOptional(secret, Password)
-	applicationCredentialID := getOptional(secret, ApplicationCredentialID)
-	applicationCredentialName := getOptional(secret, ApplicationCredentialName)
-	applicationCredentialSecret := getOptional(secret, ApplicationCredentialSecret)
-	authURL := getOptional(secret, AuthURL)
+	userName := getOptional(secret, UserName, altUserNameKey)
+	password := getOptional(secret, Password, altPasswordKey)
+	applicationCredentialID := getOptional(secret, ApplicationCredentialID, nil)
+	applicationCredentialName := getOptional(secret, ApplicationCredentialName, nil)
+	applicationCredentialSecret := getOptional(secret, ApplicationCredentialSecret, nil)
```
MartinWeindel, line 80: Maybe we should allow the DNS keys also for these three new parameters?
See https://github.com/gardener/external-dns-management/blob/master/pkg/controller/provider/openstack/handler.go#L78-L80

MartinWeindel (commented):
What about allowing `OS_APPLICATION_CREDENTIAL_xxx` key names?

pkg/openstack/credentials.go
stoyanr: Thanks, I think this was introduced recently in `external-dns-management`, for some reason I had a slightly older version and didn't see them, so I thought they are not supported for DNS for some reason. Was going to ask you about it but it slipped my mind, good that you noticed it.

stoyanr: Done (in a new commit)

stoyanr:
@MartinWeindel  Thanks for your suggestions, I addressed them, could you please take another look?

MartinWeindel (approved):
/lgtm

stoyanr:
@dkistner Rebased and ready for review once again.

pkg/openstack/client/dns.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	if rs != nil {
+		if !reflect.DeepEqual(rs.Records, records) || rs.TTL != ttl {
+			updateOpts := recordsets.UpdateOpts{
+				Records: records,
+				TTL:     &ttl,
+			}
+			_, err := recordsets.Update(c.client, zoneID, rs.ID, updateOpts).Extract()
+			return err
+		}
```
dkistner, line 61: Nit: Can we eliminate the else case by returning after the inner if block?

dkistner (approved):
Looks good to me. Thank you.

dkistner:
Sorry but needs again a rebase :( But then we can go on an merge it.
/rebase

pkg/openstack/client/dns.go
stoyanr: Done

stoyanr:
@dkistner Rebased and addressed your comments.

dkistner (approved):
/lgtm

PR #313, opened 2021-07-16
restrict app credentials to 1.19+

kon-angelo:

/area control-plane
/kind bug
/platform openstack

Using the [new application credentials](https://github.com/gardener/gardener-extension-provider-openstack/pull/300) is not supported by the in-tree openstack cloud-provider. Currently we use the in-tree provider for up to kubernetes version `v1.18.X`.  

With this PR the validation will block creation of shoots using Openstack's application credentials in unsupported versions. 

Fixes #

**Release note**:

```other operator
NONE
```


rfranzke (commented):
Can you add a unit test?

pkg/admission/validator/shoot.go
```diff
@@ -18,20 +18,21 @@ import (
...
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
+	"github.com/gardener/gardener/pkg/utils/version"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	api "github.com/gardener/gardener-extension-provider-openstack/pkg/apis/openstack"
```
dkistner, line 33: Nit: Can we move those imports from this repo up?

pkg/admission/validator/shoot.go
```diff
@@ -99,6 +100,17 @@ func (s *shoot) Validate(ctx context.Context, new, old client.Object) error {
...
 	}
 
+	// The Kubernetes version is the version of the CSI migration, where we stopped using the in-tree providers.
+	// see: pkg/webhook/controlplane/ensurer.go
+	k8sVersionLessThan19, err := version.CompareVersions(shoot.Spec.Kubernetes.Version, "<", "1.19")
+	if err != nil {
+		return err
+	}
+
+	if k8sVersionLessThan19 && s.isUsingApplicationCredentials(secret) {
```
dkistner, line 110: Should we not rather check this in `ExtractCredentials()`?
And pass `k8sVersionLessThan19` then to `ExtractCredentials()`.
Then we could also adapt the unit test there.

dkistner: We agreed not to move it to `ExtractCredentials()` as it is used in different places.

kon-angelo:
Made a second try. If this is not up to standard I can also try making the check in `ExtractCredentials`

pkg/openstack/credentials.go
```diff
@@ -119,3 +119,7 @@ func getRequired(secret *corev1.Secret, key string) (string, error) {
 	}
 	return string(value), nil
 }
+
```
dkistner, line 122: Do we need this still be as an exported function? I see it is only used in `pkg/apis/openstack/validation/shoot.go`. Can we move it there or remove the function?
If you want to keep it there then please add a doc string.

kon-angelo: I inlined the function since it's used in one place only and it's a one-liner and amply commented its usage.

dkistner (dismissed):
/lgtm

dkistner (approved):
/lgtm

PR #317, opened 2021-07-20
Deploy shoots in existing Neutron networks

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

Allows the creation of shoots into existing user-managed networks.

Fixes #

**Release note**:

```other operator
Shoots can now be deployed in existing Neutron networks. The network can be specified by its ID in the respective shoot's infrastructure configuration.
```

### DIFF
--- a/charts/internal/machineclass/templates/machineclass.yaml
+++ b/charts/internal/machineclass/templates/machineclass.yaml
@@ -37,15 +37,13 @@ providerSpec:
     availabilityZone: {{ $machineClass.availabilityZone }}
     flavorName: {{ $machineClass.machineType }}
     keyName: {{ $machineClass.keyName }}
-{{- if $machineClass.subnetID }}
-    subnetID: {{ $machineClass.subnetID }}
-{{- end }}
 {{- if $machineClass.imageID }}
     imageID: {{ $machineClass.imageID }}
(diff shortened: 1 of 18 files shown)


dkistner (commented):
Had a first look and looks overall good.
Just a few minor comments and the question how we proceed with the subnet id in the machine classes.

charts/internal/machineclass/templates/machineclass.yaml
```diff
@@ -37,6 +37,9 @@ providerSpec:
     availabilityZone: {{ $machineClass.availabilityZone }}
     flavorName: {{ $machineClass.machineType }}
     keyName: {{ $machineClass.keyName }}
+{{- if $machineClass.subnetID }}
```
dkistner, line 40: Is that check now still required when we always pass a subnet id?

dkistner (commented):
I played a little with it. Just a small comment.

Otherwise it looks good to me.

charts/internal/machineclass/templates/machineclass.yaml
kon-angelo: Nice catch

dkistner (approved):

PR #322, opened 2021-07-30
Add option ignoreVolumeAZ to allow for differences between volume and compute AZ names

gesslein:
Add option ignoreVolumeAZ to allow for differences between volume and compute AZ names

/area control-plane
/area storage
/kind enhancement
/platform openstack

Introduce CloudProfile option `ignoreVolumeAZ` which directly translates to option `ignore-volume-az` in `[BlockStorage]` section of csi config.

This option and where it is needed is described upstream as:
```
ignore-volume-az Optional. When Topology feature enabled, by default, PV volume node affinity is populated with volume accessible topology, which is volume AZ. But, some of the openstack users do not have compute zones named exactly the same as volume zones. This might cause pods to go in pending state as no nodes available in volume AZ. Enabling ignore-volume-az=true, ignores volumeAZ and schedules on any of the available node AZ. Default false.
```
https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md#block-storage

This enabled us to use gardener with an openstack that uses compute zones like `abc1-a`, `abc1-b`, `abc1-c` but the volume zone is always `abc1`.

**Release note**:

```feature operator
Add option ignoreVolumeAZ to allow for differences between volume and compute AZ names.
```


gardener-robot-ci-2:
Thank you @gesslein for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

dkistner:
Hi @gesslein,

thanks for the contribution. 
I had a first high level scan and looks so far good to me.
Will have another more detailed one beginning next week.

/needs second-opinion
/invite @kon-angelo

gesslein:
Thanks for the quick feedback.

You are right, I did not think about older versions not having that option. I did now test with a k8s 1.19.x and respectively cinder-csi-plugin v1.19.0, and it did indeed not work. But the option did no harm either.

For k8s 1.18 it is a bit puzzling, because the option does exist in the legacy cloud provider https://v1-18.docs.kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#block-storage but works differently there. I just did a quick hack to add the option to all cloudproviders.conf's and was able to schedule pods with volumes then.

To make it not so confusing, and I'm not sure 1.18 is of much relevance for new gardener adopters, I think only adding this for 1.20 and newer is the way to go. And I hope make it as clear as possible in the documentation that only 1.20.x and newer supports this.

dkistner (approved):
Ho @gesslein
sorry for the delay.
Thanks for testing the with `v1.19/v1.18` and for adding the version switch in the csi config.
I agree adopting it with `v1.20` sound good to me.

I'm now fine with it.
/lgtm

rfranzke (approved):
/lgtm

PR #334, opened 2021-09-09
Make the linter fail on missing or wrong doc comments

stoyanr:

/area quality, dev-productivity
/kind enhancement
/platform openstack

Makes the linter fail on missing or wrong doc comments and a few other common style errors, and fixes all such issues that already exist.

Fixes #

Making the linter fail on missing or wrong doc comments and a few other common style errors consists of enabling `revive` in the `golangci-lint` configuration and excluding certain `revive` errors that we are perhaps not interested in.

I have fixed the following `revive` errors (as regexes that could be included in the `golangci-lint` configuration if desired):

```
- "exported: comment on exported (var|const|type|method|function) .* should be of the form "
- "exported: exported (var|const|type|method|function) .* should have comment or be unexported"
- if-return # redundant if \.\.\.; err != nil check, just return error instead
```

I have disabled the following `revive` errors via the `golangci-lint` configuration:

```
- var-naming # ((var|const|struct field|func) .* should be .*
- dot-imports # should not use dot imports
- package-comments # package comment should be of the form
- indent-error-flow # if block ends with a return statement, so drop this else and outdent its block
- "exported: (type|func) name will be used as .* by other packages, and that stutters;"
```

See also https://github.com/gardener/gardener/pull/4627 and https://github.com/gardener/gardener-extension-provider-aws/pull/410.

**Release note**:

```other developer
Missing or wrong doc comments and a few other common style errors will now be reported by the linter.
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -5,3 +5,15 @@ run:
 linters:
   disable:
   - unused
+  enable:
+  - revive
+
+issues:
+  exclude-use-default: false
+  exclude:
+  # revive
+  - var-naming # ((var|const|struct field|func) .* should be .*
+  - dot-imports # should not use dot imports
+  - package-comments # package comment should be of the form
+  - indent-error-flow # if block ends with a return statement, so drop this else and outdent its block
+  - "exported: (type|func) name will be used as .* by other packages, and that stutters;"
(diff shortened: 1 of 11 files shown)


rfranzke (approved):
/lgtm

rfranzke:
@kon-angelo: Same as https://github.com/gardener/gardener-extension-provider-aws/pull/410#issuecomment-916919217, do you think we can get it merged?

kon-angelo:
@rfranzke After looking in more detail, they are small changes so yes it is possible. 

/unhold
/lgtm

PR #338, opened 2021-09-27
re-introduce request-timeouts for ccm

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

Allow the configuration of request-timeout for OpenStack clients used by control plane components (cloud-controller-manager and csi-driver). The configuration is done based on the respective cloudprofile value.
 
Fixes #

**Release note**:

```other operator
Allow configuration of request timeout  for control plane components (CCM, CSI) via cloudprofile
```

### DIFF
--- a/pkg/apis/openstack/types_cloudprofile.go
+++ b/pkg/apis/openstack/types_cloudprofile.go
@@ -43,7 +43,6 @@ type CloudProfileConfig struct {
 	// logical names and versions to provider-specific identifiers.
 	MachineImages []MachineImages
 	// RequestTimeout specifies the HTTP timeout against the OpenStack API.
-	// TODO: propagate timeout settings to all components (CCM,CSI,MCM,provider-extension)
 	RequestTimeout *metav1.Duration
 	// RescanBlockStorageOnResize specifies whether the storage plugin scans and checks new block device size before it resizes
 	// the filesystem.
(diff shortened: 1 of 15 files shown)


kon-angelo:
Maybe a point of discussion is making the requestTimeout from `*string` to `metav1.Duration`. Duration certainly feels like the correct type and pushes the validation towards the json unmarshalling level but probably makes reasoning about the error message harder. 

Also from my tests there is no issue with the upgrade given that the cloudprofile's `requestTimeout` already contained a valid `Duration` (which is should as a prerequisite to pass the validator check)

pkg/apis/openstack/types_cloudprofile.go
```diff
@@ -43,7 +43,8 @@ type CloudProfileConfig struct {
 	// logical names and versions to provider-specific identifiers.
 	MachineImages []MachineImages
 	// RequestTimeout specifies the HTTP timeout against the OpenStack API.
-	RequestTimeout *string
+	// TODO: propagate timeout settings to all components (CCM,CSI,MCM,provider-extension)
```
rfranzke, line 46: This TODO should not be in the API but rather in the controller, I guess :)

rfranzke (commented):
Overall it looks good

rfranzke (approved):
/lgtm

PR #354, opened 2021-11-12
Improve the error message when a required part of webhookConfig is missing

dimityrmirchev:

/area quality
/kind bug
/platform openstack

Previously when trying to install the admission helm chart without providing `.global.webhookConfig.caBundle` (for example) the helm deployment was failing with an error similar to this one:
```
... at <.Values.global.webhookConfig.caBundle>: wrong type for value; expected string; got interface {}

```

After the change the same installation will fail with a message that is actually desired.
```
... execution error at (gardener-extension-admission-openstack/charts/application/templates/validatingwebhook-validator.yaml:35:17): .Values.global.webhookConfig.caBundle is required
```

Fixes #

**Release note**:

```other operator

```


MartinWeindel (approved):
/lgtm

PR #359, opened 2021-12-01
Pass NodeTemplate to MachineClass after reading from Worker resource

himanshu-kun:

/area scalability
/kind enhancement
/platform openstack

This PR
* enables extension Openstack to read [nodeTemplate](https://github.com/gardener/gardener/blob/a408d6d9720c170b377c6476b59a5ace5fff7a6b/pkg/apis/extensions/v1alpha1/types_worker.go#L152) from worker resource and pass it onto machineClass with extra fields like InstanceType,region and zone. This is required for generic-scale-from-zero.
* allows users to provide runtime resources in `providerConfig` in shoot.yaml and then prioritizing it over cloudprofile resources while filling it in machineClass 

Fixes #
A step in [gardener/autoscaler#27](https://github.com/gardener/autoscaler/issues/27)

- will need to revendor g/g v1.37 and MCM v0.42 to access new API fields
- comments needed on the changes made  in `charts/internal/machineclass/templates/machineclass.yaml`

**Release note**:

```feature user
NONE
```


himanshu-kun:
/invite @AxiomSamarth @kon-angelo @dkistner 
Please provide early set of reviews

pkg/controller/worker/machines.go
```diff
@@ -145,10 +145,16 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
 				"region":           w.worker.Spec.Region,
 				"availabilityZone": zone,
 				"machineType":      pool.MachineType,
-				"keyName":          infrastructureStatus.Node.KeyName,
-				"networkID":        infrastructureStatus.Networks.ID,
-				"podNetworkCidr":   extensionscontroller.GetPodNetwork(w.cluster),
-				"securityGroups":   []string{nodesSecurityGroup.Name},
+				"nodeTemplate": machinev1alpha1.NodeTemplate{
+					Capacity:     pool.NodeTemplate.Capacity,
```
dkistner, line 149: Is it guaranteed that `pool.NodeTemplate` is always set mean `!= nil`?

himanshu-kun: yes , although its an optional field in the Worker API for backward compatibility , but its always set on the worker resource by gardenlet [here] (https://github.com/gardener/gardener/blob/153ebe7320e112211f05df97eabb71e34bfe8e62/pkg/operation/botanist/component/extensions/worker/worker.go#L231)

AxiomSamarth (commented):
/lgtm

pkg/controller/worker/machines.go
dkistner: Thanks for explaining.

dkistner:
Looks good to me.
We can bring that in once the dependencies are resolved.

pkg/controller/worker/machines.go
ialidzhikov: I think this field was added 8days ago and will be available in Gardener v1.37.0, right? But you should also consider cases where latest provider-openstack extensios runs against Gardener < v1.37.0. Hence, I guess you need proper handling when this field is not set (nil).

himanshu-kun: Yes @ialidzhikov , its a valid point. I'll add this field only when its not `nil` now. Thanks

rfranzke:
@himanshu-kun Please rebase this PR and mark it as 'ready'

AxiomSamarth (approved):
/lgtm

dkistner (approved):
/rebase

other than that:
/lgtm

kon-angelo (approved):
/lgtm

PR #371, opened 2022-01-05
Volume type support for root disk of VMs

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

Allows specifying the type of the root volume used in instances and passes that to the MCM extension (as implemented [here](https://github.com/gardener/machine-controller-manager-provider-openstack/pull/43))

Fixes #

**Release note**:

```other operator
Provider now supports specifying the volume type for the root disk of nodes.
```


dkistner (approved):
/lgtm

Issue #372, opened 2022-01-07
Nodes detect wrong internal IP when more than one network interface is attached 


jkmw:
**How to categorize this issue?**

/area networking
/kind bug
/platform openstack

**What happened**:

In order to reach internal services, our nodes have attached a second (private) network interface. This leads to some nodes identifying the wrong address as the internal IP. This means, for example, that logs can no longer be fetched from pods running on these nodes.

**What you expected to happen**:

The nodes should always recognise/receive the correct IP address as internal-ip, regardless of how many network interfaces are attached to the node.

**How to reproduce it (as minimally and precisely as possible)**:

1. Create a shoot-cluster
1. Connecting a second network interface to a separate internal network on all nodes
1. After a while, the `INTERNAL-IP` of some nodes will be wrong

Here is an example of the nodes before the interface was added:

```
kubectl get node -o wide

NAME                                                  STATUS   ROLES    AGE   VERSION   INTERNAL-IP    EXTERNAL-IP   OS-IMAGE           KERNEL-VERSION   CONTAINER-RUNTIME
shoot--my-project--my-cluster-default-z1-5b34b-2hnjg  Ready    <none>   92m   v1.21.3   10.250.0.147   <none>        Ubuntu 20.04 LTS   5.4.0-1015-kvm   docker://20.10.7
shoot--my-project--my-cluster-default-z1-5b34b-82b7z  Ready    <none>   90m   v1.21.3   10.250.0.132   <none>        Ubuntu 20.04 LTS   5.4.0-1015-kvm   docker://20.10.7
shoot--my-project--my-cluster-default-z1-5b34b-9btt6  Ready    <none>   89m   v1.21.3   10.250.3.29    <none>        Ubuntu 20.04 LTS   5.4.0-1015-kvm   docker://20.10.7
shoot--my-project--my-cluster-default-z1-5b34b-5hr4c  Ready    <none>   87m   v1.21.3   10.250.2.250   <none>        Ubuntu 20.04 LTS   5.4.0-1015-kvm   docker://20.10.7
```

And this is how it looks after we have attached another interface:

```
kubectl get node -o wide

NAME                                                  STATUS   ROLES    AGE   VERSION   INTERNAL-IP    EXTERNAL-IP   OS-IMAGE           KERNEL-VERSION   CONTAINER-RUNTIME
[...]

PR #373, opened 2022-01-07
Add `internalNetworkName` to cloud-provider-config

jkmw:

/area networking
/kind bug
/platform openstack

[cloud-provider-openstack](https://github.com/kubernetes/cloud-provider-openstack) offers the possibility to [identify the internal network](https://github.com/kubernetes/cloud-provider-openstack/blob/v1.22.1/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#networking). This PR adds this option to the `cloud-provider-config`

This is necessary so that nodes can identify the correct internal IP when more than one network interface is attached to them.

Fixes gardener/gardener-extension-provider-openstack#372

I have tried to find the right places in the code to make this change, but I am not sure I have found them.
Unfortunately I could not find out if `technicalID` is the best choice to determine the network name and hope for hints...

**Release note**:

```other operator
Add `internalNetworkName` to cloud-provider-config
```

### DIFF
--- a/charts/internal/cloud-provider-config/templates/_cloud-provider-config-networking.tpl
+++ b/charts/internal/cloud-provider-config/templates/_cloud-provider-config-networking.tpl
@@ -1,0 +1,6 @@
+{{- define "cloud-provider-config-networking" -}}
+[Networking]
+{{- if .Values.internalNetworkName }}
+internal-network-name="{{ .Values.internalNetworkName }}"
+{{- end }}
+{{- end -}}
--- a/charts/internal/cloud-provider-config/templates/cloud-provider-config.yaml
+++ b/charts/internal/cloud-provider-config/templates/cloud-provider-config.yaml
@@ -3,6 +3,7 @@
 {{ include "cloud-provider-config-credentials" . }}
 {{ include "cloud-provider-config-meta" . }}
 {{ include "cloud-provider-config-loadbalancer" . }}
+{{ include "cloud-provider-config-networking" . }}
 {{- end -}}
 ---
 apiVersion: v1
--- a/charts/internal/cloud-provider-config/values.yaml
+++ b/charts/internal/cloud-provider-config/values.yaml
@@ -31,6 +31,8 @@ useOctavia: false
 # - name: D
 #   floatingNetworkID: "1234"
 #   floatingSubnetTags: tag1,tag2
+# [Networking]
+# internalNetworkName: shoot--my-project--my-cluster
 # [BlockStorage]
 rescanBlockStorageOnResize: false
 ignoreVolumeAZ: false
--- a/hack/api-reference/api.md
+++ b/hack/api-reference/api.md
@@ -1087,6 +1087,17 @@ string
 </tr>
 <tr>
 <td>
+<code>name</code></br>
+<em>
+string
+</em>
+</td>
+<td>
+<p>Name is the Network name.</p>
+</td>
+</tr>
+<tr>
+<td>
 <code>floatingPool</code></br>
 <em>
 <a href="#openstack.provider.extensions.gardener.cloud/v1alpha1.FloatingPoolStatus">
--- a/pkg/apis/openstack/types_infrastructure.go
+++ b/pkg/apis/openstack/types_infrastructure.go
@@ -74,6 +74,8 @@ type NodeStatus struct {
 type NetworkStatus struct {
 	// ID is the Network id.
 	ID string
+	// Name is the Network name.
+	Name string
 	// FloatingPool contains information about the floating pool.
 	FloatingPool FloatingPoolStatus
 	// Router contains information about the Router and related resources.
--- a/pkg/apis/openstack/v1alpha1/types_infrastructure.go
+++ b/pkg/apis/openstack/v1alpha1/types_infrastructure.go
@@ -78,6 +78,8 @@ type NodeStatus struct {
 type NetworkStatus struct {
 	// ID is the Network id.
 	ID string `json:"id"`
+	// Name is the Network name.
+	Name string `json:"name"`
 	// FloatingPool contains information about the floating pool.
 	FloatingPool FloatingPoolStatus `json:"floatingPool"`
 	// Router contains information about the Router and related resources.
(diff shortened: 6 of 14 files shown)


gardener-robot-ci-1:
Thank you @jkmw for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

dkistner:
Generally looking good to me.
I just checked briefly but I guess there are no version constraints of this feature back until k8s v1.17?

/invite @kon-angelo If you have any additions?

kon-angelo:
@jkmw thank you for your contribution. 

Aside from @dkistner's question, my only other comment is related to your question regarding the use of `technicalID`. Some time ago we added a feature that allows the use of a preexisting network instead of a gardener managed one | [ref](https://github.com/gardener/gardener-extension-provider-openstack/pull/317). With that in mind, the internal network may not always correspond to the technicalID. 

Unfortunately if I understand it correctly CCM only supports the network name as an input and not the network ID which is freely available in the infrastructure status. The first solutions that come to mind is to either determine the name from the network ID in the CP reconciliation or enhancing the infrastructure status with the network name. 

WDYT?

jkmw:
You're absolutely right @kon-angelo, I hadn't considered the possibility of specifying a different network, thank you for pointing that out.

I can't think of any other option except to use gophercloud-client to get the appropriate name for the network and then use it.
I will try to make this adjustment...

jkmw:
Ok, I have at least technically succeeded in implementing the requirement. However, I had to extend an interface without knowing whether this would have an impact on any other implementations ...

dkistner:
Well I would prefer to extend the `infrastructureStatus` and add the name of the network during the infrastructure reconciliation to the status instead of looking it up in the controlplane controller.

BTW: In Openstack not all resource names are unique. What will happen when multiple network with same name exist in the same project?

The extension for the status should be rather minor.

1. Extend the infrastructure status api [ref](https://github.com/gardener/gardener-extension-provider-openstack/blob/master/pkg/apis/openstack/types_infrastructure.go#L74-L83) and [ref](https://github.com/gardener/gardener-extension-provider-openstack/blob/master/pkg/apis/openstack/v1alpha1/types_infrastructure.go#L77-L87)
2. The terraform config must output the name of the network similar like for the network id [ref](https://github.com/gardener/gardener-extension-provider-openstack/blob/master/pkg/internal/infrastructure/templates/main.tpl.tf#L142-L144).
3. In addition we need to construct the infrastructure status correctly [ref](https://github.com/gardener/gardener-extension-provider-openstack/blob/master/pkg/internal/infrastructure/terraform.go#L210-L270).

Then we could simply get the network name in the controlplane controller from the status.

kon-angelo:
> BTW: In Openstack not all resource names are unique. What will happen when multiple network with same name exist in the same project?

What usually is happening with CCM in these cases is that it cannot resolve the correct network and starts throwing errors in the reconciliation.
But what I saw (at least in current master) is that for this field it is simply a string comparison that it is used in, so I would expect this to work. AFAIK, CCM just checks the server interface's network name with the provided list. 

> Well I would prefer to extend the infrastructureStatus and add the name of the network during the infrastructure reconciliation to the status instead of looking it up in the controlplane controller.
No objection, just stating here that doing this for existing clusters the name would not be available until the next infrastructure reconciliation (where the name will be populated).

dkistner:
> No objection, just stating here that doing this for existing clusters the name would not be available until the next infrastructure reconciliation (where the name will be populated).

Yes a full reconciliation of the Shoot would be required to make use of it, but I think that is acceptable. Better then always doing additional unnecessary calls (we can fetch the information as we anyway have it present in the infrastructure reconciliation).

dkistner:
> What usually is happening with CCM in these cases is that it cannot resolve the correct network and starts throwing errors in the reconciliation.
But what I saw (at least in current master) is that for this field it is simply a string comparison that it is used in, so I would expect this to work. AFAIK, CCM just checks the server interface's network name with the provided list.

Can we validate/proof this? I would also expect that occm throws errors in that case. I just wanted to be sure that there are no unforeseen side effects.

jkmw:
The PR has been adjusted again. I thought it was a good idea to use the technicalID as a fallback in case of an empty network name.

What do you think?

kon-angelo:
> Can we validate/proof this? I would also expect that occm throws errors in that case. I just wanted to be sure that there are no unforeseen side effects.

The only use of the `internalNetworkName` for CCM i found is [this](https://github.com/kubernetes/cloud-provider-openstack/blob/9b8de65448e7872ac4e83c7342e498127155253b/pkg/openstack/instances.go#L539). We could can do a hands-on test for that.

> The PR has been adjusted again. I thought it was a good idea to use the technicalID as a fallback in case of an empty network name.

Based on the CCM snippet above and considering a case for a cluster using an external network this change would still cause problems as I think CCM will have trouble discerning node addresses and that may break existing shoots.

If in case no name is provided in the infrastructure status you do not pass this option to CCM, then existing shoots remain as is and shoots that need this feature can trigger an infrastructure reconciliation. Disclaimer: I didn't try the new version to determine exactly the effect on CCM for existing shoots(that are using externally managed network) but it would be interesting to know that.

jkmw:
> Based on the CCM snippet above and considering a case for a cluster using an external network this change would still cause problems as I think CCM will have trouble discerning node addresses and that may break existing shoots.

Good point, now the `internal-network-name` is also only written if it could be extracted from the status. Otherwise it remains empty.

dkistner (dismissed):
@jkmw The PR looks now good to me. Thanks for applying the suggestions.

@kon-angelo could you please have another look?

/lgtm
/needs second-opinion

kon-angelo (approved):
/lgtm

kon-angelo:
@jkmw The PR looks to me too. It is currently failing our CI checks due to a couple unformatted files. Could you take a look and update accordingly ?

jkmw:
Ok, there was one blank line too much

dkistner (approved):

kon-angelo (approved):
/lgtm

PR #375, opened 2022-01-11
Use separate resolv.conf for kubelet

MartinWeindel:

/area control-plane
/kind enhancement
/platform openstack

Debian derivates like Garden Linux and Ubuntu are using `systemd-resolved` for  network name resolution. This service also contains a full-featured local resolver at `127.0.0.53`.
With this PR, the kubelet uses a separate resolve config file `/etc/resolv-for-kubelet.conf` linked to `/run/systemd/resolve/resolv.conf`. If the cloud profile defines `resolvConfOptions`, the kubelet will use a patched copy of `/run/systemd/resolve/resolv.conf` as already introduced with PR #342. As the kubelet configuration is defined globally for all worker pools, the implementation is a little complicated. It needs to provide the separate `/etc/resolv-for-kubelet.conf` even if the OS is has no `systemd-resolved`. The `update-resolv-conf.service` is dealing with this case.
A separate resolve config file for the kubelet is useful on OS with `systemd-resolved`. The standard `/etc/resolv.conf` can then reference the standard local resolver. 
Besides, the (harmless) duplicate name server issue seen on Garden Linux disappears with these changes.

As these changes have impact for all used operating systems, it has been validated for Garden Linux, SUSE chost, and Ubuntu (both with and without resolvConfOptions in the cloud profile)

**Release note**:

```other operator
Use separate resolv.conf for kubelet (optionally patched with resolvConfOptions from the cloud profile)
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -505,8 +505,7 @@ func (e *ensurer) EnsureAdditionalFiles(ctx context.Context, gctx gcontext.Garde
 	if err != nil {
 		return err
 	}
-	options := getResolveConfOptions(cloudProfileConfig)
-	e.addAdditionalFilesForResolvConfOptions(options, new)
+	e.addAdditionalFilesForResolvConfOptions(getResolveConfOptions(cloudProfileConfig), new)
 	return nil
 }
 
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -419,6 +419,9 @@ func (e *ensurer) EnsureKubeletConfiguration(ctx context.Context, gctx gcontext.
 		new.FeatureGates[csiMigrationCompleteFeatureGate] = true
 	}
 
+	// resolv-for-kubelet.conf is created by update-resolv-conf.service
+	new.ResolverConfig = "/etc/resolv-for-kubelet.conf"
+
 	return nil
 }
 
@@ -458,17 +461,11 @@ func (e *ensurer) EnsureKubeletCloudProviderConfig(ctx context.Context, gctx gco
 
 // EnsureAdditionalUnits ensures that additional required system units are added.
 func (e *ensurer) EnsureAdditionalUnits(ctx context.Context, gctx gcontext.GardenContext, new, old *[]extensionsv1alpha1.Unit) error {
-	cloudProfileConfig, err := getCloudProfileConfig(ctx, gctx)
-	if err != nil {
-		return err
-	}
-	if len(getResolveConfOptions(cloudProfileConfig)) > 0 {
-		e.addAdditionalUnitsForResolvConfOptions(new)
-	}
+	e.addAdditionalUnitsForResolvConfOptions(new)
 	return nil
 }
 
-// addAdditionalUnitsForResolvConfOptions installs a systemd service to update `resolv.conf`
+// addAdditionalUnitsForResolvConfOptions installs a systemd service to update `resolv-for-kubelet.conf`
 // after each change of `/run/systemd/resolve/resolv.conf`.
 func (e *ensurer) addAdditionalUnitsForResolvConfOptions(new *[]extensionsv1alpha1.Unit) {
 	var (
@@ -480,7 +477,7 @@ PathChanged=/run/systemd/resolve/resolv.conf
 WantedBy=multi-user.target
 `
 		customUnitContent = `[Unit]
-Description=add options to /etc/resolv.conf after every change of /run/systemd/resolve/resolv.conf
+Description=update /etc/resolv-for-kubelet.conf on start and after each change of /run/systemd/resolve/resolv.conf
 After=network.target
 StartLimitIntervalSec=0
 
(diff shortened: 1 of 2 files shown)


marwinski (dismissed):
see my comments. I assume it would work but I would change some things in future commits. 

- we really do need this for all cloud providers 
- this probably should be generic and more or less OS independent. In this case we have systemd-resolved systems and non-systemd-resolved systems. A script can handle this in a generic way (which it does but it could be a bit cleaner)

This said, the PR looks good to me but I'd propose some improvements in the future.

pkg/webhook/controlplane/ensurer.go
```diff
@@ -522,20 +519,48 @@ func (e *ensurer) addAdditionalFilesForResolvConfOptions(options []string, new *
 		template          = `#!/bin/sh
 
 file=/run/systemd/resolve/resolv.conf
-tmp=/etc/resolv.conf.new
-dest=/etc/resolv.conf
+tmp=/etc/resolv-for-kubelet.conf.new
+dest=/etc/resolv-for-kubelet.conf
 line=%q
 
+# use separate resolv conf for kubelet based on /run/systemd/resolve/resolv.conf
```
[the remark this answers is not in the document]
MartinWeindel: good idea, changed the script

pkg/webhook/controlplane/ensurer.go
```diff
@@ -522,20 +519,48 @@ func (e *ensurer) addAdditionalFilesForResolvConfOptions(options []string, new *
...
+  ln -s /etc/resolv.conf "$tmp"
+fi
+mv "$tmp" "$dest"
+echo updated "$dest"
+
+file2=/run/systemd/resolve/stub-resolv.conf
+tmp2=/etc/resolv.conf.new
+dest2=/etc/resolv.conf
+
+# use link to /run/systemd/resolve/stub-resolv.conf as /etc/resolv.conf
```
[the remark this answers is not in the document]
MartinWeindel: Removed any change of the `/etc/resolv.conf`

pkg/webhook/controlplane/ensurer.go
```diff
@@ -508,9 +505,8 @@ func (e *ensurer) EnsureAdditionalFiles(ctx context.Context, gctx gcontext.Garde
 	if err != nil {
 		return err
 	}
-	if options := getResolveConfOptions(cloudProfileConfig); len(options) > 0 {
-		e.addAdditionalFilesForResolvConfOptions(options, new)
-	}
+	options := getResolveConfOptions(cloudProfileConfig)
+	e.addAdditionalFilesForResolvConfOptions(options, new)
```
dkistner, line 509:
```suggestion
	e.addAdditionalFilesForResolvConfOptions(getResolveConfOptions(cloudProfileConfig), new)
```
Nit

pkg/webhook/controlplane/ensurer.go
```diff
@@ -521,21 +517,41 @@ func (e *ensurer) addAdditionalFilesForResolvConfOptions(options []string, new *
...
+{
+    if [ -f /run/systemd/resolve/resolv.conf ]; then
+      return 0
+    else
+      return 1
+    fi
+}
+
+rm -f "$tmp"
+if is_systemd_resolved_system; then
```
dkistner, line 534: Why not using `[ -f /run/systemd/resolve/resolv.conf ]` as condition here directly?
Is it the function `is_systemd_resolved_system()` somewhere as used?

pkg/webhook/controlplane/ensurer.go
```diff
@@ -522,20 +519,48 @@ func (e *ensurer) addAdditionalFilesForResolvConfOptions(options []string, new *
...
-  echo "$line" >> "$tmp"
-  mv "$tmp" "$dest"
+  if [ "$line" = "" ]; then
+    rm "$tmp"
+    ln -s "$file" "$tmp"
+  else
+    cp "$file" "$tmp"
+    echo "" >> "$tmp"
+    echo "# updated by update-resolv-conf.service (installed by gardener-extension-provider-openstack)" >> "$tmp"
+    echo "$line" >> "$tmp"
```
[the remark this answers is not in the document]
dkistner: @marwinski Does that harm (e.g. multiple nameserver configurations)?

pkg/webhook/controlplane/ensurer.go
```diff
@@ -521,21 +517,41 @@ func (e *ensurer) addAdditionalFilesForResolvConfOptions(options []string, new *
...
+      return 1
+    fi
+}
+
+rm -f "$tmp"
+if is_systemd_resolved_system; then
+  if [ "$line" = "" ]; then
+    ln -s /run/systemd/resolve/resolv.conf "$tmp"
+  else
+    cp /run/systemd/resolve/resolv.conf "$tmp"
```
dkistner, line 538: We could copy `/run/systemd/resolve/resolv.conf` always to `$tmp` and only append `$line` if not empty. wdyt?
But it's just a nit.

dkistner (commented):
Looks generally good to me.
Just a few minor suggestions (nothing mandatory).

pkg/webhook/controlplane/ensurer.go
MartinWeindel: It was suggested by my other reviewer (see https://github.com/gardener/gardener-extension-provider-openstack/pull/375#discussion_r782951877). 
And I like it as it makes the semantics more visible.

pkg/webhook/controlplane/ensurer.go
MartinWeindel: I think it is easier to reason about the configuration with the symbolic link. Otherwise you have to compare the file content.

dkistner: Hmm I don't get this, but as said this was just a nit. I'm also fine with this approach.

dkistner (approved):
/lgtm
/needs second-opinion
/ping @marwinski

marwinski (approved):

PR #378, opened 2022-01-17
Update RBAC for leader election of `machine-controller-manager`

shafeeqes:

/area control-plane
/kind enhancement
/platform openstack

Update RBAC for leader election of `machine-controller-manager` before vendoring MCM `v0.43.0`.

Fixes part of gardener/gardener#4742

Related to gardener/machine-controller-manager#662

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/machine-controller-manager/seed/templates/clusterrole.yaml
+++ b/charts/internal/machine-controller-manager/seed/templates/clusterrole.yaml
@@ -26,3 +26,19 @@ rules:
   - events
   verbs:
   - "*"
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - create
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  resourceNames:
+  - machine-controller-manager
+  verbs:
+  - get
+  - watch
+  - update


kon-angelo (approved):
/lgtm

charts/internal/machine-controller-manager/seed/templates/clusterrole.yaml
```diff
@@ -26,3 +26,19 @@ rules:
...
+  resources:
+  - leases
+  verbs:
+  - create
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  resourceNames:
+  - machine-controller-manager
```
shafeeqes, line 40: we probably need `machine-controller` resourceName here too. I missed it.
@kon-angelo

shafeeqes: We can add it while we update the image for the machine-controller-manager-provider-openstack too.

kon-angelo: You mean to make a new release of mcm-provider-openstack with the v0.43 MCM changes ?

No objection, but if it is something that can be fixed immediately with an extra PR, I am not sure if we should wait for the mcm-provider-openstack release.

shafeeqes: But only with the mcm-provider-openstack release with revendored MCM v0.43.0, we need rbac for the `machine-controller` resource.
I can open a PR in gardener-extension-provider-openstack once there is a new release of the same.

PR #383, opened 2022-01-21
Revendor `gardener/gardener@v1.39`

rfranzke:
**How to categorize this PR?**

/area security
/kind enhancement
/merge squash

**What this PR does / why we need it**:
Revendor `gardener/gardener@v1.39`.

**Special notes for your reviewer**:
✅ ~Depends on #364, hence, PR is in draft state.~
✅ ~Depends on `gardener/gardener@v1.39` release, hence, PR is in draft state.~

**Release note**:

```noteworthy operator
NONE
```

PR #384, opened 2022-01-22
Kubernetes Version Configurability per Worker Pool

rfranzke:

/area usability
/kind enhancement
/platform openstack
/merge squash

This PR implements support for `gardener/gardener`'s `WorkerPoolKubernetesVersion` feature gate. As explained in https://github.com/gardener/gardener/issues/3501#issuecomment-1009061189, it
- adds validation to prevent using kubelet versions lower than the CSI migration version (note that this ensures that the feature is only usable once the cluster with its control plane and **all** its worker pools was upgraded to at least the CSI migration version).
- adapts the `controlpane` webhook to use the effective kubelet version when computing the CSI migration feature gate names.

Part of gardener/gardener#3501

✅ ~Depends on #383, hence, PR is in draft state.~

**Release note**:

```feature operator
This extension does now support `gardener/gardener`'s `WorkerPoolKubernetesVersion` feature gate, i.e., having [worker pools with overridden Kubernetes versions](https://github.com/gardener/gardener/blob/8a9c88866ec5fce59b5acf57d4227eeeb73669d7/example/90-shoot.yaml#L69-L70).
```
```feature user
In case `gardener/gardener`'s `WorkerPoolKubernetesVersion` feature gate is enabled, it's possible having [worker pools with overridden Kubernetes versions](https://github.com/gardener/gardener/blob/8a9c88866ec5fce59b5acf57d4227eeeb73669d7/example/90-shoot.yaml#L69-L70) for `Shoot`s whose `.spec.kubernetes.version` is greater or equal than the CSI migration version (`1.19`).
```

### DIFF
--- a/docs/usage-as-end-user.md
+++ b/docs/usage-as-end-user.md
@@ -205,3 +205,8 @@ Every OpenStack shoot cluster that has at least Kubernetes v1.19 will be deploye
 It is compatible with the legacy in-tree volume provisioner that was deprecated by the Kubernetes community and will be removed in future versions of Kubernetes.
 End-users might want to update their custom `StorageClass`es to the new `cinder.csi.openstack.org` provisioner.
 Shoot clusters with Kubernetes v1.18 or less will use the in-tree `kubernetes.io/cinder` volume provisioner in the kube-controller-manager and the kubelet.
+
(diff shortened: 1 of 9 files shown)


dkistner (approved):
/lgtm

PR #388, opened 2022-02-01
Update `csi-driver-cinder` for kubernetes v1.23

acumino:

/area storage
/kind enhancement
/platform openstack

This PR adds `csi-driver-cinder` for Kubernetes v1.23 to the extension.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -68,7 +68,12 @@ images:
   sourceRepository: github.com/kubernetes/cloud-provider-openstack
   repository: docker.io/k8scloudprovider/cinder-csi-plugin
   tag: "v1.22.0"
-  targetVersion: ">= 1.22"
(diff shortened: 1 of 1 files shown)


charts/images.yaml
```diff
@@ -69,6 +69,11 @@ images:
   repository: docker.io/k8scloudprovider/cinder-csi-plugin
   tag: "v1.22.0"
   targetVersion: ">= 1.22"
```
rfranzke, line 71: This should be changed to `1.22.x`

rfranzke (changes requested):

rfranzke (approved):
/lgtm

charts/images.yaml
acumino: Thanks, Updated it.

PR #408, opened 2022-03-03
enable storage-classes config via cloud-profile

rgroemmer:

/area storage
/kind enhancement
/platform openstack

**What this PR does / why we need it:**

The PR enables support for multiple volume-types in OpenStack, which can be set via cloud-profile.

**Release note**:

```feature operator
StorageClasses can now be set via cloudprofile to support e.g. multiple volume-types.
```


gardener-robot-ci-2:
Thank you @rgroemmer for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

kon-angelo:
Hello @rgroemmer. Your contribution is very much appreciated. It's true that we were missing such customization hook and I can see how it would be beneficial. I would like to touch on some topics regarding the implementation.

This is an implementation for a storage class rendered. The storage class as a k8s resource is generic and it doesn't contain any provider specific information (sans the *contents* of the parameters). Since there is anything tying it to openstack, I would say that is seems to be more fitting inside the core gardener `CloudProfile` resource. This would enable it also for other cloud providers that would care to take advantage of this feature.

The second thing I want to touch is the interaction with the default storageclasses that we currently deploy. With this change we will depend entirely on the `cloudprofile` for `storageClass`es forgoing any default we currently have.  I wonder if a combination between default classes and additional ones coming from the `cloudProfile` would make sense.

dergeberl:
Hi @kon-angelo, 
I am not sure if it is better to have it direct in Cloudprofile. In the most cases there is no need to change sc for cloud providers like GCP or AWS. If you like it globally in the Cloudprofile we can open a PR on gardener to add it there. The implementation from this PR should be the same except the field which are used for the setting.

For the second point. It is implemented that if no sc is set in the cloudprofile a default one (like the old one) will be created. I think it should be possible to don`t have any default sc because in some cases this SC would not work.

kon-angelo:
Hi @dergeberl. There are also other infrastructure providers in the wild that could benefit from making this change in `gardener/gardener` even if it is not as useful for the major cloud providers.

The way you handle the storageClass (templates, annotations etc.) is great and my only point of contention is the source of the data. I propose that you open an issue in `gardener/gardener` and we can further discuss if the storageClass can fit in the core `CloudProfile` there with more audience.   

Regarding the deployment semantics, I agree with you. Ideally, you would be able to specify between defaults, cloudprofile or none but I consider the current semantics (Cloudprofile if possible, otherwise use the defaults) the best alternative. As it is, you can only have 2 possible states. Maybe an `enable` type variable or removing the `omitempty` can be used to encode the additional state.

kon-angelo:
Hi @dergeberl. There are also other infrastructure providers in the wild that could benefit from making this change in `gardener/gardener` even if it is not as useful for the major cloud providers.

The way you handle the storageClass (templates, annotations etc.) is great and my only point of contention is the source of the data. I propose that you open an issue in `gardener/gardener` and we can further discuss if the storageClass can fit in the core `CloudProfile` there with more audience.   

Regarding the deployment semantics, I agree with you. Ideally, you would be able to specify between defaults, cloudprofile or none but I consider the current semantics (Cloudprofile if possible, otherwise use the defaults) the best alternative. As it is, you can only have 2 possible states. Maybe an `enable` type variable or removing the `omitempty` can be used to encode the additional state.

charts/internal/shoot-storageclasses/templates/storageclasses.yaml
```diff
@@ -1,24 +1,30 @@
...
-apiVersion: {{ include "storageclassversion" . }}
-kind: StorageClass
-metadata:
-  name: default
-  annotations:
+    {{- if $value.default }}
     storageclass.kubernetes.io/is-default-class: "true"
-    resources.gardener.cloud/delete-on-invalid-update: "true"
+    {{- else }}
+    storageclass.kubernetes.io/is-default-class: "false"
```
dkistner, line 13: Can we not simply omit this block in case the storageclass shouldn't be default?

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -457,17 +458,94 @@ func (vp *valuesProvider) GetControlPlaneShootCRDsChartValues(
...
+			} else {
+				allSc[i]["provisioner"] = "cinder.csi.openstack.org"
+			}
+			if sc.ReclaimPolicy != nil && *sc.ReclaimPolicy != "" {
+				allSc[i]["reclaimPolicy"] = sc.ReclaimPolicy
+			}
+			if sc.VolumeBindingMode != nil && *sc.VolumeBindingMode != "" {
+				allSc[i]["volumeBindingMode"] = sc.VolumeBindingMode
+			}
+		}
```
dkistner, line 509:
```suggestion
		for i, sc := range providerConfig.StorageClasses {
		       var storageClassValues = map[string]interface{}{
		          "name": sc.Name,
		       }
		
			if sc.Default != nil && *sc.Default {
				storageClassValues["default"] = true
			}
			
			if len(sc.Annotations) != 0 {
				storageClassValues["annotations"] = sc.Annotations
			}
			if len(sc.Labels) != 0 {
				storageClassValues["labels"] = sc.Labels
			}
			if len(sc.Parameters) != 0 {
				storageClassValues["parameters"] = sc.Parameters
			}
			
			storageClassValues["provisioner"] = "cinder.csi.openstack.org"
			if sc.Provisioner != nil && *sc.Provisioner != "" {
			    storageClassValues["provisioner"]  = sc.Provisioner
			}
			
			if sc.ReclaimPolicy != nil && *sc.ReclaimPolicy != "" {
				storageClassValues["reclaimPolicy"] = sc.ReclaimPolicy
			}
			
			if sc.VolumeBindingMode != nil && *sc.VolumeBindingMode != "" {
				storageClassValues["volumeBindingMode"] = sc.VolumeBindingMode
			}
			
			allSc[i] = storageClassValues
		}
```

Just an optional structural suggestion as I found it a bit cleaner.

dkistner (commented):
I guess for now we can move on by making storageclasses configurable via the CloudProfile only for the Openstack extension. If there is a need for this on other infrastructures we can later follow up on https://github.com/gardener/gardener/issues/5644

We should ensure a fallback to the intree storageclasses in case the cloudprofile does not contain any storage class configurations.

charts/internal/shoot-storageclasses/templates/storageclasses.yaml
dergeberl: Yes this should be possible. I think it is more like a styling topic. I like the visibility of the default state. But I can remove it if you like.

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -457,17 +458,94 @@ func (vp *valuesProvider) GetControlPlaneShootCRDsChartValues(
...
+			},
+				{
+					"name":              "default-class",
+					"provisioner":       "cinder.csi.openstack.org",
+					"volumeBindingMode": bindMode,
+				},
+			},
+		}
+	}
+	return values, nil
```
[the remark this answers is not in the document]
dergeberl: I changed your code suggested code a bit (it is not possible to access a `map[string]interface`)

I used the existing `consts` from kubernetes and created 2 for the openstack values.

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -457,17 +458,90 @@ func (vp *valuesProvider) GetControlPlaneShootCRDsChartValues(
...
+		{
+			"name":              "default-class",
+			"provisioner":       openstack.CSIStorageProvisioner,
+			"volumeBindingMode": storagev1.VolumeBindingImmediate,
+		}}
+
+	if k8sVersionLessThan119 {
+		if k8sVersionLessThan112 {
+			storageclasses[0]["volumeBindingMode"] = storagev1.VolumeBindingImmediate
+			storageclasses[1]["volumeBindingMode"] = storagev1.VolumeBindingWaitForFirstConsumer
```
dkistner, line 536: For all the defaults that we currently have we use always `WaitForFirstConsumer` as binding mode, now we switch to `Immediate`. As this change affects only newly created volumes I guess we should switch for all defaults to `Immediate`.
@kon-angelo wdyt?

dergeberl:
@dkistner is there something I can do from my side or is it ready to merge?

dkistner (approved):
Looks good to me.
Let's wait for @kon-angelo for a final look then we move on.

dergeberl:
@dkistner @kon-angelo 
Can someone have a look if we can merge this PR?

pkg/controller/controlplane/valuesprovider.go
kon-angelo: I would also prefer to have it as `WaitForFirstConsumer` to avoid having the volumes to be created in the wrong zone.

kon-angelo:
@dergeberl Looks great. Only minor nit is that comment from @dkistner about the binding mode

kon-angelo (dismissed):
/lgtm

I tested it and seems to work fine. I still want to change the default classes to use `WaitForFirstConsumer` but we can do this in subsequent PR.

kon-angelo:
@dergeberl can you make the last fixes in the PR and then we can proceed with merging.

You need to rerun `make revendor` and possibly check the formatting for some files due to the newer version of the linter used.

pkg/controller/controlplane/valuesprovider.go
dergeberl: I changed all to `WaitForFirstConsumer` and also removed the now unused check for older k8s versions.

dergeberl:
> @dergeberl can you make update the PR to pass the verify check?
> 
> You need to rerun `make revendor` and possibly check the formatting for some files due to the newer version of the linter used.

Should be fine now. I also rebased to the latest master to avoid conflicts in the `go.mod`.

kon-angelo:
@dergeberl linter check is still failing

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/gophercloud/utils v0.0.0-20200204043447-9864b6f1f12f
 	github.com/onsi/ginkgo/v2 v2.1.4
 	github.com/onsi/gomega v1.20.0
+	github.com/pkg/errors v0.9.1
```
ialidzhikov, line 21: Why you now introduce this when there was explicit PR to clean up this dependency? Ref https://github.com/gardener/gardener-extension-provider-openstack/pull/331

dkistner: Yes, we should remove `github.com/pkg/errors` again. (Easy to happen for long running PRs like this).

PR #412, opened 2022-03-08
Update manifest.yaml

Kostov6:
Absolute to relative path conversion in .docforge/manifest.yaml for simplification

None

**Release note**:

```doc dependency
Paths transformations in .docforge/manifest.yaml for simplification
```

### DIFF
--- a/.docforge/manifest.yaml
+++ b/.docforge/manifest.yaml
@@ -3,4 +3,4 @@ structure:
   source: /README.md
 - name: docs
   nodesSelector:
-      path: /docs
+    path: /docs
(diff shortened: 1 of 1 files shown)


.docforge/manifest.yaml
```diff
@@ -1,13 +1,6 @@
...
 - name: docs
-  nodes:
-  - nodesSelector:
-      path: https://github.com/gardener/gardener-extension-provider-openstack/tree/{{$defaultBranch}}/docs
-links:
-  downloads:
-    scope:
-      "gardener/gardener-extension-provider-openstack/(blob|raw)/(.*)/docs": ~
+  nodesSelector:
+      path: /docs
```
ialidzhikov, line 6: This does not look like a properly formatted yaml:
```suggestion
  nodesSelector:
    path: /docs
```

ialidzhikov (dismissed):

.docforge/manifest.yaml
Kostov6: yes, I added a fixing commit. Thanks for noting

PR #416, opened 2022-03-11
Allow projected volumes in the csi-driver-node PSP

ialidzhikov:
/kind bug
/platform openstack

Similar to https://github.com/gardener/gardener-extension-provider-aws/pull/510.

We were not able to reproduce similar PSP issue for K8s >= 1.19 but even though that I think it still makes sense to keep the PSP up-to-date and to have it reflecting the needed permissions.

**Release note**:

```other operator
NONE
```


rfranzke (approved):

kon-angelo (commented):
/lgtm

PR #420, opened 2022-03-21
configure parameters for csi components

nschad:
Signed-off-by: Niclas Schad <niclas.schad@stackit.de>

/kind enhancement
/kind api-change

Optional config for csi components

Fixes #

**Release note**:
```other operator
Add possibility to configure csi components (csi-attacher, csi-resizer, ....)
```


gardener-robot-ci-1:
Thank you @nschad for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

nschad:
Do you need me to rebase now or shall I wait for initial review first?

pkg/apis/config/types.go
```diff
@@ -45,6 +47,78 @@ type ETCD struct {
...
+	RetryIntervalMax *string
+	// ReconcileSync Resync frequency of the attached volumes with the driver.
+	ReconcileSync *string
+	// Timeout Timeout of all calls to the container storage interface driver.
+	Timeout *string
+	// Verbose The verbosity level.
+	Verbose *string
+}
+
+type CSIDriverCinder struct {
```
dkistner, line 80: Many of those structs are exactly the same. How about embed the things that they have in common e.g. `Verbose` or `Timeout`?

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -750,12 +751,139 @@ func getCSIControllerChartValues(
...
+
+func getCSIProvisionerArgs(provisioner *config.CSIProvisioner) map[string]interface{} {
+	csiProvisioner := make(map[string]interface{})
+	csiProvisionerArgs := make(map[string]interface{})
+	if provisioner.Verbose != nil {
+		csiProvisionerArgs["verbosity"] = provisioner.Verbose
+	}
+	if provisioner.Timeout != nil {
+		csiProvisionerArgs["timeout"] = provisioner.Timeout
+	}
```
dkistner, line 806: Once we embed these fields in the api we can move this in a dedicated function and just call it in each `getCSI...Args()` func.

dkistner (commented):
Generally I'm fine with making the csi components more configurable.
The way you propose it would make the csi component configuration so to say globally for a Gardener installations. Do you see any reason to override these settings for individual clusters?

nschad:
> Generally I'm fine with making the csi components more configurable. The way you propose it would make the csi component configuration so to say globally for a Gardener installations. Do you see any reason to override these settings for individual clusters?

We feel like having the config globally should be more than enough due to it being "infrastructure-related"

pkg/apis/config/types.go
nschad: Yeah I didn't really like the solution either tbh. I guess we could try and identify common args but these could also change. And the fact that we have like 6-7 different csi components doesn't make it necessarily easier. Would be cool if we could have some super generic option where u just enter what you want like `extraArgs` helm-chart kind off stuff.

nschad: Is that something you would be interested in @dkistner ?

dkistner:
I agree it should be infrastructure/installation globally and not decided on cluster level.
Will soon have another look.

pkg/apis/config/types.go
dkistner: Well at the end I guess we need something like this:

```go
type CSI struct {
	// CSIAttacher is the configuration for the csi-attacher
	CSIAttacher *CSIAttacher
	// CSIDriverCinder is the configuration for the csi-driver-cinder
	CSIDriverCinder *CSIBaseArgs
	// CSIProvisioner is the configuration for the csi-provisioner
	CSIProvisioner *CSIBaseArgs
	// CSIResizer is the configuration for the csi-resizer
	CSIResizer *CSIBaseArgs
	// CSISnapshotController is the configuration for the csi-shapshot-controller
	CSISnapshotController *CSIBaseArgs
	// CSISnapshotter is the configuration for the csi-snapshotter
	CSISnapshotter *CSIBaseArgs
	// CSILivenessProbe is the configuration for the csi-liveness-probe
	CSILivenessProbe *CSIBaseArgs
}

type CSIAttacher struct {
        CSIBaseArgs
	// RetryIntervalStart The exponential backoff for failures.
	RetryIntervalStart *string
	// RetryIntervalMax The exponential backoff maximum value.
	RetryIntervalMax *string
	// ReconcileSync Resync frequency of the attached volumes with the driver.
	ReconcileSync *string
}

type CSIBaseArgs struct {
	// Timeout Timeout of all calls to the container storage interface driver.
	Timeout *string
	// Verbose The verbosity level.
	Verbose *string
}
```

I guess with something like this we can avoid a lot DRY also for the helm values generation in the controlplane controller.

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/deployment-csi-driver-controller.yaml
```diff
@@ -93,15 +90,17 @@ spec:
...
         - --leader-election
         - --leader-election-namespace=kube-system
-        - --timeout={{ .Values.timeout }}
-        - --v=5
+        {{- if .Values.csiProvisioner.extraArgs.timeout }}
+        - --timeout={{ .Values.csiProvisioner.extraArgs.timeout }}
+        {{- end }}
+        {{- if .Values.csiProvisioner.extraArgs.verbosity }}
+        - --v={{ .Values.csiProvisioner.extraArgs.verbosity }}
+        {{- end }}
```
nschad, line 98: Instead of this we could also do:
```
            {{- range $key, $value := .Values.csiProvisioner.extraArgs }}
            - "-{{ $key }}={{ $value }}"
            {{- end }}
````

Same for all the other components obviously, would get rid of some further dry code.

**Comment by dkistner** [comment]
@nschad thanks for removing/reducing the dry code in this PR.
I guess the PR is now in a good state to move on. But a rebase would be required.
/rebase

File: charts/internal/seed-controlplane/charts/csi-driver-controller/templates/deployment-csi-driver-controller.yaml (hunk 3 above)
Comment (nschad) [reply charts/internal/seed-controlplane/charts/csi-driver-controller/templates/deployment-csi-driver-controller.yaml:98]: @dkistner Do you want me to add the loop?

**Comment by nschad** [comment]
Rebased. Let me know if you want me to add the comment suggestion from may

File: charts/internal/seed-controlplane/charts/csi-driver-controller/values.yaml [hunk 4]
```diff
@@ -12,46 +12,74 @@ images:
...
-  driver:
+csiDriverCinder:
+  resources:
     requests:
       cpu: 20m
       memory: 50Mi
     limits:
-      memory: 360Mi
-  provisioner:
+      cpu: 50m
```
dkistner, line 23: Please adopt here the old memory limits and drop the cpu limits.
We recently changed them to more accurate values. The values here would revert that.

pkg/apis/config/types.go
```diff
@@ -47,6 +49,64 @@ type ETCD struct {
...
+	CSIBaseArgs
+}
+
+type CSIResizer struct {
+	CSIBaseArgs
+}
+
+type CSISnapshotController struct {
+	CSIBaseArgs
+}
```
dkistner, line 101: We don't need those structs as they just embed `CSIBaseArgs`.
Could you please use those in the `CSI` struct as suggested?
 https://github.com/gardener/gardener-extension-provider-openstack/pull/420/files#r879148760

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -668,12 +673,128 @@ func getCSIControllerChartValues(
...
+		}
+		if csi.CSILivenessProbe != nil {
+			values["csiLivenessProbe"] = getCSILivenessProbeArgs(csi.CSILivenessProbe)
+		}
+		if csi.CSIProvisioner != nil {
+			values["csiProvisioner"] = getCSIProvisionerArgs(csi.CSIProvisioner)
+		}
+		if csi.CSISnapshotter != nil {
+			values["csiSnapshotter"] = getCSISnapshotterArgs(csi.CSISnapshotter)
+		}
```
dkistner, line 698: If we adapt the `CSI` struct to use the `CSIBaseArgs` for those field we can also just use the `getCSIBaseArgs()` function with different inputs. Only the `CSIAttacher` needs its own `get...Args()` func and can call internally `getCSIBaseArgs()`.

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/deployment-csi-driver-controller.yaml
dkistner: Yes, would reduce also some dry code.

pkg/controller/controlplane/valuesprovider.go
nschad: Alright I tried to shrink the code as much as possible.

nschad:
Hi @dkistner,

Thank you for all the feedback. What if we keep it super generic and leave it for the User to configure the fields correctly. By changing the `CSI` struct to something like this

```go
type CSI struct {
	// CSIAttacher is the configuration for the external-attacher
	CSIAttacher *map[string]string
	// CSIDriverCinder is the configuration for the csi-driver-cinder
	CSIDriverCinder *map[string]string
	// CSIProvisioner is the configuration for the external-provisioner
	CSIProvisioner *map[string]string
	// CSIResizer is the configuration for the external-resizer
	CSIResizer *map[string]string
	// CSISnapshotController is the configuration for the snapshot-controller
	CSISnapshotController *map[string]string
	// CSISnapshotter is the configuration for the external-snapshotter
	CSISnapshotter *map[string]string
	// CSILivenessProbe is the configuration for the livenessprobe
	CSILivenessProbe *map[string]string
}
```
Then we truly have a super **dynamic configuration** which requires **no** code changes when wanting to adjust something. It also comes with some disadvantages (no validation, no code-completion, etc.) obviously but it's the only solution I see without having to create a **PR** everytime you wan't to add a CLI Flag to a CSI component

dkistner:
Well I tend to disagree. If we simply pass any kind of value we have no control over the csi config. What would happen for instance for cases where we already forseen some flags. I guess we don't want to make them configurable. This can easily break.

charts/internal/seed-controlplane/charts/csi-driver-controller/templates/deployment-csi-driver-controller.yaml
```diff
@@ -117,15 +116,14 @@ spec:
         - --kubeconfig=/var/run/secrets/gardener.cloud/shoot/generic-kubeconfig/kubeconfig
         - --leader-election
         - --leader-election-namespace=kube-system
-        - --timeout={{ .Values.timeout }}
```
dkistner, line 120: Sorry my bad for this. But I guess with this generic loop it can happen that we don't have any flag for verbosity or timeout nor default values for them. I guess we should only make `extraArgs` configurable that we know that they can be configured via the `cloudprofile`.

same for the other csi components. wdyt?

nschad: > But I guess with this generic loop it can happen that we don't have any flag for verbosity or timeout nor default values for them.

That is true. It would be on the user to decide if he wants to differ from the csi component default not on gardener anymore.

> I guess we should only make extraArgs configurable that we know that they can be configured via the cloudprofile.

Can't really follow. Isn't this what I proposed with the map[string]string?

timebertt:
We're reworking this PR to simplify it and keep it focused on the configuration we actually need

PR #421, opened 2022-03-21
Add ready check for Webhook server

acumino:

/area open-source
/kind enhancement
/platform openstack

The PR adds ready check for webhook server. The ready check wait for 10 sec for webhook server to be ready, otherwise throws error.

Part of https://github.com/gardener/gardener/issues/4772

**Release note**:

```other operator
NONE
```


cmd/gardener-extension-provider-openstack/app/app.go
```diff
@@ -212,6 +215,18 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
...
 
+			if err := mgr.AddReadyzCheck("informer-sync", gardenerhealthz.NewCacheSyncHealthz(mgr.GetCache())); err != nil {
+				return fmt.Errorf("could not add readycheck for informers: %w", err)
+			}
+
+			if err := mgr.AddHealthzCheck("ping", healthz.Ping); err != nil {
+				return fmt.Errorf("could not add health check to manager: %w", err)
+			}
+
+			if err := mgr.AddReadyzCheck("webhook-server", mgr.GetWebhookServer().StartedChecker()); err != nil {
```
dkistner, line 226: Also just for confirmation the readyness probe will be ok when the webhook server of the manager are up and the informers are synced?

cmd/gardener-extension-provider-openstack/app/app.go
```diff
@@ -212,6 +215,18 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("could not add controllers to manager: %w", err)
 			}
 
+			if err := mgr.AddReadyzCheck("informer-sync", gardenerhealthz.NewCacheSyncHealthz(mgr.GetCache())); err != nil {
+				return fmt.Errorf("could not add readycheck for informers: %w", err)
+			}
+
+			if err := mgr.AddHealthzCheck("ping", healthz.Ping); err != nil {
```
dkistner, line 222: If I recall correctly this `healthz.Ping` returns always nil, which means the health check/liveness probe will always be true, right?

charts/gardener-extension-provider-openstack/templates/deployment.yaml
```diff
@@ -53,6 +53,9 @@ spec:
         - --webhook-config-server-port={{ .Values.webhookConfig.serverPort }}
         - --disable-controllers={{ .Values.disableControllers | join "," }}
         - --disable-webhooks={{ .Values.disableWebhooks | join "," }}
+        {{- if .Values.healthPort }}
```
dkistner, line 56: Will `.Values.healthPort` not always set? There is a default value in the `values.yaml` defined.

cmd/gardener-extension-provider-openstack/app/app.go
acumino: Yes.

charts/gardener-extension-provider-openstack/templates/deployment.yaml
acumino: Since `health-bind-address` has a default value, it can be skipped from here.

cmd/gardener-extension-provider-openstack/app/app.go
acumino: Yes

charts/gardener-extension-provider-openstack/templates/deployment.yaml
dkistner: I guess you mean the default value you set in the `values.yaml` or the one which you define in the code? As far as I see they are both the same. My point that the if clause wouldn't be required here, when the chart has a default value, right? Same for the admission controller. But it is just a nit.

acumino: My idea was if someone uses the same yamls but don't give the value of `healthPort` in values.yaml in that case it will throw an error, though not the case in our scenario. What's you call? Should I remove the `if`?

dkistner: The `.Values.healthPort` will not be empty as you define a value for it in the `values.yaml` of the chart. This can be overridden in case someone want so provide own value for it, but your default value will be there.  As said it does not hurt to have the `if` but it's not required in my opinion. 
So up to you I'm fine with keeping or removing it as it is just a nit.

dkistner (approved):
/lgtm

charts/gardener-extension-provider-openstack/templates/deployment.yaml
acumino: Remove it in c2a041cf94eed49c66d41676f5a911163e6b59bf. PTAL

dkistner (approved):
/lgtm

dkistner (approved):
/lgtm

PR #425, opened 2022-03-24
remove --external-hostname flag from kube-apiserver to succeed k8s conformance check

### LINKED ISSUE #366: K8s Conformance check failing for 1.21

/area control-plane
/kind bug
/platform openstack

**What happened**:

Running the conformance check on a Shoot with Kubernetes version 1.21, we had the following failed test:

```
[sig-auth] ServiceAccounts should support OIDC discovery of service account issuer
# Error output:
2021/12/10 10:46:15 OK: Got token
2021/12/10 10:46:15 validating with in-cluster discovery
2021/12/10 10:46:15 OK: got issuer https://api.shootname.internal.x.x
2021/12/10 10:46:15 Full, not-validated claims:
openidmetadata.claims{Claims:jwt.Claims{Issuer:"https://api.shootname.internal.x.x", Subject:"system:serviceaccount:svcaccounts-3132:default", Audience:jwt.Audience{"oidc-discovery-test"}, Expiry:1639133774, NotBefore:1639133174, Iss
2021/12/10 10:46:15 OK: Constructed OIDC provider for issuer https://api.shootname.internal.x.x
2021/12/10 10:46:15 failed to validate with in-cluster discovery: failed to verify signature: fetching keys oidc: get keys failed Get "https://x.x.x.x:443/openid/v1/jwks": x509: certificate is valid for 127.0.0.1, 100.64.0.1, not x.x.x.x
2021/12/10 10:46:15 falling back to validating with external discovery
2021/12/10 10:46:15 OK: got issuer https://api.shootname.internal.x.x
2021/12/10 10:46:15 Full, not-validated claims:
openidmetadata.claims{Claims:jwt.Claims{Issuer:"https://api.shootname.internal.x.x", Subject:"system:serviceaccount:svcaccounts-3132:default", Audience:jwt.Audience{"oidc-discovery-test"}, Expiry:1639133774, NotBefore:1639133174, Iss
2021/12/10 10:46:15 Get "https://api.shootname.internal.x.x/.well-known/openid-configuration": x509: certificate signed by unknown authority
```

This problem should be solved by Gardener by setting `--external-hostname` to the API Server URL for OIDC config:

https://github.com/gardener/gardener/blob/5867624a45a1cd92e6fd48e1fae5773b67bc8ee5/pkg/operation/botanist/component/kubeapiserver/deployment.go#L453

But the `--external-hostname` is set to the IP address of the API server by the Openstack extension webhook:

https://github.com/gardener/gardener-extension-provider-openstack/blob/719a343a36b133764f8563b6ec1eec25dfe6d6c1/pkg/webhook/controlplaneexposure/ensurer.go#L83

Which leads to the failing test. We fixed it by removing the line 83 in the Openstack extension webhook ensurer, so the Gardener external hostname setting will be used.

Should we remove this line in a PR or is this needed for other functionality?

We would expect the API URL of the API Server to be passed as the external hostname.

**Environment**:

- Gardener version (if relevant): 1.36
- Extension version: 1.22
- Kubernetes version (use `kubectl version`): 1.21.5

dergeberl:
/area certification
/kind bug
/platform openstack

If `ManagedIstio` and `APIServerSNI` are disabled the k8s conformance check fails. More details see: #366 

Fixes #366

**Release note**:
```breaking operator
Fix failing k8s conformance check if ManagedIstio and APIServerSNI are disabled
```

### DIFF
--- a/pkg/webhook/controlplaneexposure/ensurer.go
+++ b/pkg/webhook/controlplaneexposure/ensurer.go
@@ -80,7 +80,6 @@ func (e *ensurer) EnsureKubeAPIServerDeployment(ctx context.Context, gctx gconte
 
 	if c := extensionswebhook.ContainerWithName(new.Spec.Template.Spec.Containers, "kube-apiserver"); c != nil {
 		c.Command = extensionswebhook.EnsureStringWithPrefix(c.Command, "--advertise-address=", address)
-		c.Command = extensionswebhook.EnsureStringWithPrefix(c.Command, "--external-hostname=", address)
 	}
 	return nil
 }


pkg/webhook/controlplaneexposure/ensurer.go
```diff
@@ -80,7 +80,6 @@ func (e *ensurer) EnsureKubeAPIServerDeployment(ctx context.Context, gctx gconte
 
 	if c := extensionswebhook.ContainerWithName(new.Spec.Template.Spec.Containers, "kube-apiserver"); c != nil {
 		c.Command = extensionswebhook.EnsureStringWithPrefix(c.Command, "--advertise-address=", address)
-		c.Command = extensionswebhook.EnsureStringWithPrefix(c.Command, "--external-hostname=", address)
```
dkistner, line 83: @ialidzhikov, @ScheererJ Do you have an idea on the impact for the SNI case?

ref: https://github.com/gardener/gardener-extension-provider-openstack/issues/366#issuecomment-996698944

ScheererJ: @dkistner As mentioned by @ialidzhikov in #366, this should have no impact whatsoever on the SNI case as the logic jumps out of the function in line 63 due to `v1beta1helper.IsAPIServerExposureManaged(new)` returning `true` in case SNI is enabled.

I am not sure why the external hostname is overwritten here. As far as I can tell the external hostname is always set by gardener since 1.33. Maybe in the past, the certificates had the IPs as subject and not the DNS name.

ScheererJ (dismissed):
/lgtm

ialidzhikov:
I recall we had a discussion about removal of this flag from provider-azure and provider-opensack in https://github.com/gardener/gardener/pull/4067 in the context of https://github.com/gardener/gardener/issues/4023. I have not enough context anymore but I think we already wanted to remove the flag from the corresponding extensions. Maybe @dimityrmirchev you have more than me as author of #4067 and can comment on this PR.

pkg/webhook/controlplaneexposure/ensurer.go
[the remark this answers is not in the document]
dergeberl: Oh yes, i missed that. I fixed the failing tests.

ialidzhikov (approved):
/lgtm
/squash

dimityrmirchev:
AFAIK the `--external-hostname` is only used to generate the `jwks` uri of an apiserver. Kubernetes documentation states that is also used for generating Swagger API Docs but I am not sure if this even affects Gardener in any way. A quick fetch of the `openapi` docs located at `/openapi/v2` shows that the output does not include any occurrences of the `external-hostname`. As already mentioned this change should not affect the SNI scenario. And in the case SNI is not enabled I think it is best to promote the internal/external apiserver's uri since the certificate provided is valid for these DNS names.

kon-angelo (approved):
/lgtm

PR #427, opened 2022-03-31
Add snaphot validation webhook

acumino:

/area open-source
/kind enhancement
/platform openstack

This PR adds the webhook to provide additional validation to volume snapshot objects i.e. VolumeSnapshot and VolumeSnapshotContent.

Similar to https://github.com/gardener/gardener-extension-provider-aws/pull/507

**Release note**:

```feature user
The provider-openstack extension now installs the external-snapshotter's validating webhook server for VolumeSnapshot and VolumeSnapshotContent objects. For more details check the corresponding [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1900-volume-snapshot-validation-webhook#kep-1900-add-additional-validation-to-volume-snapshot-objects).
```


charts/internal/seed-controlplane/charts/csi-driver-controller/templates/csi-snapshot-validation-webhook-vpa.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: VerticalPodAutoscaler
+metadata:
+  name: csi-snapshot-webhook-vpa
+  namespace: {{ .Release.Namespace }}
+spec:
+  resourcePolicy:
+    containerPolicies:
+    - containerName: csi-snapshot-validation
+      minAllowed:
+        memory: 25Mi
```
dkistner, line 11: Could you please use the memory requests of the webhook controller defined in the `values.yaml` for the `minAllowed.memory`?

`.Values.csiSnapshotValidationWebhook.resources.memory`

pkg/controller/healthcheck/add.go
```diff
@@ -79,6 +91,12 @@ func RegisterHealthChecks(mgr manager.Manager, opts healthcheck.DefaultAddArgs)
 				HealthCheck:   general.NewSeedDeploymentHealthChecker(openstack.CSISnapshotControllerName),
 				PreCheckFunc:  csiEnabledPreCheckFunc,
 			},
+			// TODO(acumino): Enable this health check in v1.27.
+			{
+				ConditionType: string(gardencorev1beta1.ShootControlPlaneHealthy),
+				HealthCheck:   general.NewSeedDeploymentHealthChecker(openstack.CSISnapshotValidation),
+				PreCheckFunc:  healthCheck,
+			},
```
kon-angelo, line 99: This should be commented out right ?

kon-angelo (changes requested):

pkg/controller/healthcheck/add.go
acumino: No, the Precheck function is doing that only, it is skipping the health check.

kon-angelo: Then maybe I don't understand the TODO. The precheck ia already only executed if the `csi-snapshot-validation` is present so what would need to be changed come v1.27. Can you elaborate ?

acumino: We can't start the health check for this deployment with the release of it, because it will make the shoot status to unhealthy because of the missing snapshot-validation deployment. the status can remain false until the shoot is reconciled, to don't populate our landscapes dashboard with this false positive we skip the health check with the first release and it will be enabled with the upcoming release.

kon-angelo (approved):
/lgtm

PR #450, opened 2022-05-20
Managed Application Credentials

### LINKED ISSUE #429: Manged application credentials
/area security
/kind enhancement
/platform openstack

Add support for application credentials managed by the openstack extension for the user provided technical users.
These application credentials will be used to interact with openstack apis instead of the technical users.

**Why is this needed**:
This will avoid disruptions while rotating users on Openstack, so that the Openstack extension and its managed components (ccm, csi, mcm) can still communicate with openstack apis even if the user credentials have changed.

dkistner:
/area security
/kind enhancement
/platform openstack

This PR adds support for application credentials managed by the Openstack extension.
The application credentials will be owned, created and deleted by the (technical) user which is provided by the Shoot owner.
All api calls to the Openstack layer will be made by the application credential, except the ones that are required to manage the application credential itself.

Using a managed application credential will help to not disrupt a cluster while the (technical) user of the cluster is exchanged, expired or its secret is rotated.

The application credential will be renewed after a configured lifetime during the next Shoot reconciliation/deletion operation. Even if the extension is unable the manage the application credential the application credential will be invalidated after a configured period via Openstack means.
The managed application credential usage is disabled by default and can be enabled via the controller config of the extension.

Fixes #429

The following tasks need to be done fulfilled before this PR can be merged:
- [x] Feature documentation
- [ ] Unit test coverage
- [ ] End-to-end tests completed

The PR is opened as draft to start gathering feedback.
/invite @kon-angelo 
/cc @donistz 

**Release note**:

```feature user
The `gardener-openstack-extension` can now manage an application credential which is used to call the Openstack layer instead of the (technical) user provided by the Shoot owner. The (technical) user will only be used to manage the application credential all other api calls to the Openstack layer will be done with the application credential. This will help to keep the Shoot functional when the technical user or its secret is rotated.
```

/squash


pkg/internal/managedappcredential/manager_ensure.go
```diff
@@ -0,0 +1,199 @@
...
+			return m.removeApplicationCredentialStore(ctx, appCredential.secret)
+		}
+
+		if oldParentUserUsable {
+			if err := m.runGarbageCollection(ctx, oldParentUser, "", true); err != nil {
+				return err
+			}
+		}
+
+		return m.removeApplicationCredentialStore(ctx, appCredential.secret)
```
kon-angelo, line 92: Maybe:
```
if appCredentialExists && oldParentUserUsable {
			if err := m.runGarbageCollection(ctx, oldParentUser, "", true); err != nil {
				return err
			}
}
return m.removeApplicationCredentialStore(ctx, appCredential.secret)

File: pkg/internal/managedappcredential/manager_ensure.go [hunk 2]
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	// If the new parent user is itself an application credential, then no managed
+	// application credential can be used. It will be tried to clean up with the
+	// old user if this one is usuable.
+	if newParentUser.isApplicationCredential() {
+		if appCredentialExists {
```
kon-angelo, line 99: Do you need `oldParentUsable` check here too ? It seems similar to my previous comment so can they be consolidated e .g 

```
if !m.config.Enabled  || newParentUser.isApplicationCredential()
```

pkg/internal/managedappcredential/manager_ensure.go
```diff
@@ -0,0 +1,199 @@
...
+		newAppCredential, err := m.createApplicationCredential(ctx, newParentUser)
+		if err != nil {
+			return err
+		}
+
+		return m.storeApplicationCredential(ctx, newAppCredential, newParentUser)
+	}
+
+	return m.storeApplicationCredential(ctx, appCredential, newParentUser)
+}
```
kon-angelo, line 155: I get the feeling that the code can be condensed a bit more e.g.
```
if parentChanged || m.isExpired(appCredential) || !appCredentialExists {
		// Try to delete the old application credential owned by the old user.
		// This might not work as the information about this user could be stale,
		// because the user credentials are rotated, the user is not associated to
		// Openstack project anymore or it is deleted.
		if parentChanged{
				if err := oldParentUser.identityClient.DeleteApplicationCredential(ctx, oldParentUser.id, appCredential.id); err != nil {
			m.logger.Error(err, "could not delete application credential as the owning user has changed and information about owning user might be stale")
		         }
                }

		newAppCredential, err := m.createApplicationCredential(ctx, newParentUser)
		if err != nil {
			return err
		}
		if err :=  m.storeApplicationCredential(ctx, newAppCredential, newParentUser){
                    return err
                 }
}
if err := m.runGarbageCollection(ctx, newParentUser, appCredential.id, false); err != nil {
		return err
}

```
It is easy to miss a case anyway so I am not sure how big of an improvement it is.

pkg/internal/managedappcredential/manager.go
```diff
@@ -0,0 +1,68 @@
...
+		config:                 config,
+		finalizerName:          finalizerName,
+		identifier:             identifier,
+		logger:                 logger,
+		namespace:              namespace,
+		openstackClientFactory: openstackClientFactory,
+	}
+}
+
+func (m *Manager) runGarbageCollection(ctx context.Context, parent *parent, inUseAppCredentialID string, deleteInUseAppCredential bool) error {
```
kon-angelo, line 44: From the uses in `actuator_delete` and `actuator_reconcile` it is not clear to me if we need both inUseAppCredentialID and deleteInUseAppCredential. The general idea is to not delete the current in-use things and also have a mode where you delete everything. Isn't the same if you pass an empty in-use credential to delete everything ?

pkg/controller/infrastructure/add.go
```diff
@@ -38,13 +39,15 @@ type AddOptions struct {
 	// DisableProjectedTokenMount specifies whether the projected token mount shall be disabled for the terraformer.
 	// Used for testing only.
 	DisableProjectedTokenMount bool
+	// AppCredentialConfig specifies the configuration for application credential usage.
+	AppCredentialConfig controllerconfig.ApplicationCredentialConfig
 }
 
 // AddToManagerWithOptions adds a controller with the given AddOptions to the given manager.
 // The opts.Reconciler is being set with a newly instantiated actuator.
 func AddToManagerWithOptions(mgr manager.Manager, options AddOptions) error {
```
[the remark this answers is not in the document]
dkistner: Hmm not sure about this. The `github.com/gardener/gardener/extensions/pkg/controller/cmd.Switch()` gets a receiver for the `AddToManager` function and later call it just by passing the manager to it. Here we would need to pass the `AddOptions` as well in our case the `DefautAddOptions`.

pkg/internal/managedappcredential/manager_ensure.go
dkistner: Nice. Make sense.

pkg/internal/managedappcredential/manager.go
dkistner: No we don't need both. It would be sufficient to pass the `appCredentialID` as a `*string` and check for nil. That's how I had it in the beginning and then changed to current version for better readability. Now I'm rather with you and changed is back.

pkg/controller/infrastructure/add.go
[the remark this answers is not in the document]
kon-angelo: I misunderstood how it is used. The code is fine as it is

dkistner:
For the moment we will not continue with this PR. The extension and dependent components (mcm, ccm, csi etc.) are capable to work with application credentials. The lifecycle of the application credentials need for now be managed externally.

/close

PR #452, opened 2022-05-26
admission-openstack: Adapt Secrets webhook to rely on the provider label

ialidzhikov:
/area robustness
/area cost
/kind enhancement
/platform openstack

This PR:
- Adapts the Secrets webhook to rely on the provider label (similar to https://github.com/gardener/gardener-extension-provider-gcp/pull/396). With this the Secrets webhook:
  1) no longer intercepts every Secret UPDATE request but only requests for Secrets with label `provider.shoot.gardener.cloud/openstack="true"`
  2) no longer needs to list Shoots (hence, no longer has a cache for Shoots based on Informers) to understand whether a Secret is in-use by a Shoot from the same provider type ("openstack")
- Adds a new SecretBinding webhook (similar to https://github.com/gardener/gardener-extension-provider-gcp/pull/428). This is to prevent creating SecretBindings that refer Secret from wrong provider type and hence block update request for this Secret as the admission will fail because the Secret does not contain valid credentials.

Part of https://github.com/gardener/gardener-extension-provider-gcp/issues/143 but for admission-openstack

Unlike admission-{gcp,aws,azure} for admission-openstack we cannot drop the cache for SecretBindings as the Shoot validator has to get the SecretBinding -> then the Secret to fetch the domainName from it - the domainName is later used for validation.

https://github.com/gardener/gardener-extension-provider-openstack/blob/1d8ffd8d1b983b9eae12a16d5502258374a3e242/pkg/admission/validator/shoot.go#L216-L236

https://github.com/gardener/gardener-extension-provider-openstack/blob/1d8ffd8d1b983b9eae12a16d5502258374a3e242/pkg/admission/validator/shoot.go#L120-L122

**Release note**:

```breaking operator
This version of admission-openstack requires the SecretBinding provider controller to be enabled - enabled by default for gardener-controller-manager >= 1.42 or can be enabled via the gardener-controller-manager component config.
```

```other operator
The Secrets webhook of admission-openstack:
- no longer intercepts every Secret UPDATE request but only requests for Secrets that are associated with a SecretBinding with `provider.type=openstack`.
- no longer needs to list Shoots (hence, no cache for Shoots)
```

```other operator
The admission-openstack component introduces a new SecretBinding validator. It validates requests for SecretBindings and checks whether the SecretBinding refers to a valid OpenStack Secret.
```


rfranzke (approved):
/lgtm

PR #497, opened 2022-09-27
change node affinity of extensionn controller

RaphaelVogel:

/area control-plane
/kind enhancement
/platform openstack

Extension controller pod with multiple replicas is deployed with pod anti affinity `requiredDuringSchedulingIgnoredDuringExecution`. This can prevent the deployment to a seed cluster having only one node. Changing it to `preferredDuringSchedulingIgnoredDuringExecution` will allow scheduling the pod to seeds with only one node.  

Fixes #

**Release note**:

```other operator

```

### DIFF
--- a/charts/gardener-extension-provider-openstack/templates/deployment.yaml
+++ b/charts/gardener-extension-provider-openstack/templates/deployment.yaml
@@ -103,7 +103,7 @@ spec:
         {{- end }}
       affinity:
         podAntiAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
+          preferredDuringSchedulingIgnoredDuringExecution:
           - labelSelector:
               matchExpressions:
               - key: {{ include "labels.app.key" . }}


charts/gardener-extension-provider-openstack/templates/deployment.yaml
```diff
@@ -103,7 +103,7 @@ spec:
         {{- end }}
       affinity:
         podAntiAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
+          preferredDuringSchedulingIgnoredDuringExecution:
```
ialidzhikov, line 106: This could regress the HA mode of the extension that also serves webhooks - needed for GEP-20.
cc @timuthy @unmarshall

timuthy (commented):
> This can prevent the deployment to a seed cluster having only one node.

I assume that such seeds are usually used for dev and testing, so I wondered why someone can't simply deploy the extension with `.Values.replicaCount: 1`?

charts/gardener-extension-provider-openstack/templates/deployment.yaml
timuthy: As @ialidzhikov said, this is not compatible with what is desired in an HA scenario where we want to forcefully distribute the pods across nodes/zones.
If you want to support this soft affinity with multiple replicas, then I suggest to enhance the chart values by an optional affinity setting that can be passed.
Even better, use https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/ instead of `podAntiAffinity`.

unmarshall: I agree TSC would be a better option here.

PR #504, opened 2022-10-12
support switching of network overlay mode

DockToFuture:

/area networking
/kind enhancement
/platform openstack

Overlay networks for newly created calico shoot clusters are dropped now by default. To keep the overlay network `networking.providerConfig.overlay.enabled: true` needs to be set explicitly in the shoot spec.

Fixes #

**Release note**:

```other operator
Overlay networks for newly created calico shoot clusters are dropped now by default.  To keep the overlay network `networking.providerConfig.overlay.enabled: true` needs to be set explicitly in the shoot spec.
```


DockToFuture:
This PR is on hold to until a new calico release was created with https://github.com/gardener/gardener-extension-networking-calico/pull/215.

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/coreos/go-systemd/v22 v22.3.2
 	github.com/gardener/etcd-druid v0.12.3
 	github.com/gardener/gardener v1.56.0
+	github.com/gardener/gardener-extension-networking-calico v1.26.1-0.20221012083616-51b857cae1cc
```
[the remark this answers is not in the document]
DockToFuture: will update it

pkg/admission/mutator/shoot.go
```diff
@@ -0,0 +1,91 @@
...
+	if old == nil && networkConfig.Overlay == nil {
+		networkConfig.Overlay = overlay
+	}
+
+	if old != nil && networkConfig.Overlay == nil {
+		oldShoot, ok := old.(*gardencorev1beta1.Shoot)
+		if !ok {
+			return fmt.Errorf("wrong object type %T", old)
+		}
+		oldNetworkConfig, err := s.decodeNetworkingConfig(oldShoot.Spec.Networking.ProviderConfig)
```
kon-angelo, line 68: There is a very specific edge case here that blocks shoot deletion. For the current shoots where the networkingConfig is `nil`, you will still try to mutate them but on deletion updates to these fields are not possible. I think you should ignore changes to shoots with termination timestamp

charts/internal/cloud-provider-config/templates/_cloud-provider-config-route.tpl
```diff
@@ -0,0 +1,6 @@
+{{- define "cloud-provider-config-route" -}}
+[Route]
+{{- if .Values.routerID }}
+router-id="{{ .Values.routerID }}"
```
kon-angelo, line 4: There is also one more failure mode which is particular to openstack: 

Because now we do configure the cloud routes, there are rare cases where the CCM is terminated before getting to clear all the routes during shoot termination. Because there are still custom routes in the router, the deletion of the router cannot proceed.

You can see that for other cloud providers e.g. gcp ([ref](https://github.com/gardener/gardener-extension-provider-gcp/blob/9992d3829cdea7071c9c50318c17debad3969677/pkg/controller/infrastructure/actuator_delete.go#L132)) we do have custom cleaning logic that makes sure that this doesn't happen.

kon-angelo:
We talked with @DockToFuture and decided that we will merge this PR as is and address the point of the route cleaning logic in a separate PR.

kon-angelo (approved):
/lgtm

PR #511, opened 2022-11-04
allow passthrough of worker labels to VM metadata

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

With this PR the labels in a worker pool will be used to annotate the OpenStack server instance (in addition to the node object - that already being done). Because in OpenStack the metadata key is more restrictive than K8s we need to transform the key by replacing illegal characters.

Fixes #

**Release note**:

```other operator
Allow adding the `Labels` of a worker pool to the corresponding OpenStack Instance.
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -18,10 +18,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
-
-	api "github.com/gardener/gardener-extension-provider-openstack/pkg/apis/openstack"
-	"github.com/gardener/gardener-extension-provider-openstack/pkg/apis/openstack/helper"
-	"github.com/gardener/gardener-extension-provider-openstack/pkg/openstack"
+	"regexp"
 
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/controller/worker"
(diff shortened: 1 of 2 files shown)


pkg/controller/worker/machines.go
```diff
@@ -232,3 +234,15 @@ func (w *workerDelegate) generateWorkerPoolHash(pool extensionsv1alpha1.WorkerPo
...
 }
+
+// NormalizeLabelsForMachineClass because metadata in OpenStack resources do not allow for certain characters that present in k8s labels e.g. "/",
+// normalize the label by replacing illegal characters with "-"
+func NormalizeLabelsForMachineClass(in map[string]string) map[string]string {
+	notAllowedChars := regexp.MustCompile(`[^a-zA-Z0-9-_:. ]`)
+	res := make(map[string]string)
+	for k, v := range in {
+		newKey := notAllowedChars.ReplaceAllLiteralString(k, "-")
+		res[newKey] = v
```
dkistner, line 245: Just the one question: Do we also need to normalize the values?

kon-angelo: I didn't see the need for it. The values need to follow k8s conventions (enforced by gardener validation) since they are used for node labeling too and I didn't see any issue with that character set when I tried to use it as values, so I would say no.

dkistner (approved):

PR #514, opened 2022-11-17
delete infrastructure routes in case ccm is not there

DockToFuture:

/area networking
/kind enhancement
/platform openstack

All infrastructure routes within the workers node range are deleted during the deletion process. This lets the deletion still succeed when the cloud controller manager is not there anymore.

Fixes #

**Release note**:

```other operator
All infrastructure routes within the workers node range are deleted during the deletion process. This lets the deletion still succeed when the cloud controller manager is not there anymore.
```

### DIFF
--- a/pkg/controller/infrastructure/actuator_delete.go
+++ b/pkg/controller/infrastructure/actuator_delete.go
@@ -17,16 +17,20 @@ package infrastructure
 import (
 	"context"
 	"fmt"
+	"time"
 
+	api "github.com/gardener/gardener-extension-provider-openstack/pkg/apis/openstack"
 	"github.com/gardener/gardener-extension-provider-openstack/pkg/apis/openstack/helper"
 	"github.com/gardener/gardener-extension-provider-openstack/pkg/internal"
 	"github.com/gardener/gardener-extension-provider-openstack/pkg/internal/infrastructure"
 	"github.com/gardener/gardener-extension-provider-openstack/pkg/openstack"
+	openstackclient "github.com/gardener/gardener-extension-provider-openstack/pkg/openstack/client"
 	"github.com/go-logr/logr"
 
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/terraformer"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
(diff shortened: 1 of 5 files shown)


pkg/controller/infrastructure/actuator_delete.go
```diff
@@ -64,9 +68,65 @@ func (a *actuator) Delete(ctx context.Context, log logr.Logger, infra *extension
...
-		SetEnvVars(internal.TerraformerEnvVars(infra.Spec.SecretRef, credentials)...).
-		Destroy(ctx)
+	tf = tf.InitializeWith(ctx, terraformer.DefaultInitializer(a.Client(), terraformFiles.Main, terraformFiles.Variables, terraformFiles.TFVars, stateInitializer)).SetEnvVars(internal.TerraformerEnvVars(infra.Spec.SecretRef, credentials)...)
+
+	configExists, err := tf.ConfigExists(ctx)
+	if err != nil {
+		return err
+	}
+
+	vars, err := tf.GetStateOutputVariables(ctx, infrastructure.TerraformOutputKeyRouterID)
```
kon-angelo, line 89: This is an interesting choice to use the TF methods to get the router ID and I want to make sure that I understand the implications.

I am concerned with the following scenario
- creating a shoot with invalid credentials and want to delete it afterwards. In this case the state may or may not be there. The router variable will also certainly not be there. From a glance I think that your calls later e.g. `router, err := client.GetRouterByID(routerID)`  would permanently block deletion

I also think that the condition at line 103 `DoIf(configExists)` is not entirely correct. The config may exists but a routerID not necessarily. 

Maybe we should use the infrastructureStatus for the routerID ? That way you would know whether there is a router in the first place and you can use the ID directly instead of making the lookup

DockToFuture: It's similar to what we do on gcp https://github.com/gardener/gardener-extension-provider-gcp/blob/eb851f716e45336b486f3aaf46268859de2adecb/pkg/controller/infrastructure/actuator_delete.go#L62

DockToFuture: It's all done in the `ExtractTerraformState` function.

DockToFuture: If there is no routerID I will assume that there are no routes to delete and go on with the tf.Destory.

PR #521, opened 2022-11-28
do not configure cloud routes if overlay is used

kon-angelo:

/area control-plane
/kind bug
/platform openstack

Disable configuring cloud routes when an ovelay network is used. While this doesn't cause an issue for the cluster (P2P communication) but if you are reusing routers for multiple shoots with the same pod ranges then CCM instances will compete with each other and causing API overhead.

Fixes #

**Release note**:

```other operator
Disable configuring cloud routes when overlay network is used.
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -20,6 +20,10 @@ import (
 	"path/filepath"
 	"strings"
 
+	calicov1alpha1 "github.com/gardener/gardener-extension-networking-calico/pkg/apis/calico/v1alpha1"
+	"github.com/gardener/gardener-extension-networking-calico/pkg/calico"
+	"github.com/gardener/gardener/pkg/apis/core/v1beta1"
+
 	api "github.com/gardener/gardener-extension-provider-openstack/pkg/apis/openstack"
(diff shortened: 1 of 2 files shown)


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -442,7 +451,10 @@ func getConfigChartValues(
 		// detect internal network.
 		// See https://github.com/kubernetes/cloud-provider-openstack/blob/v1.22.1/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#networking
 		"internalNetworkName": infraStatus.Networks.Name,
-		"routerID":            infraStatus.Networks.Router.ID,
+	}
+
+	if !isUsingOverlay {
+		values["routerID"] = infraStatus.Networks.Router.ID
```
DockToFuture, line 457: Will the deletion still work if the routerID is not there anymore? Because we made use of it in the delete operation in the actuator.

MartinWeindel (approved):
/lgtm

pkg/controller/controlplane/valuesprovider.go
MartinWeindel: The deletion uses the router.id returned from Terraformer. The terraformer still has the router id input (or the router is created). From my understanding of the code it is unrelated to the dropping of the routerID from the cloud provider config.

PR #527, opened 2022-11-29
delete infrastructure routes in case ccm is not there

DockToFuture:

/area networking
/kind enhancement
/platform openstack

All infrastructure routes within the workers node range are deleted during the deletion process. This lets the deletion still succeed when the cloud controller manager is not there anymore.

Fixes #

**Release note**:

```other operator
All infrastructure routes within the workers node range are deleted during the deletion process. This lets the deletion still succeed when the cloud controller manager is not there anymore.
```


kon-angelo (approved):
/lgtm

PR #529, opened 2022-12-01
update calico version to v1.27.1

DockToFuture:

/area networking
/kind bug
/platform openstack

Update calico version to `v1.27.1`.

This is required because in previous calico versions the `backend` field was not omitted when set to `nil`.
This update will avoid setting `backend: null` in the shoot spec when the shoot mutator webhook in the infrastructure extension is called which leads to an error during validation.

```
error: error validating "STDIN": error validating data: unknown object type "nil" in Shoot.spec.networking.providerConfig.backend; if you choose to ignore these errors, turn validation off with --validate=false
```

Fixes #

**Release note**:

```other operator
Update calico version to `v1.27.1`.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/coreos/go-systemd/v22 v22.3.2
 	github.com/gardener/etcd-druid v0.12.3
 	github.com/gardener/gardener v1.59.0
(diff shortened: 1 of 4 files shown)


kon-angelo (approved):
/lgtm

PR #543, opened 2022-12-14
Skip network mutation when shoot specs are equal

DockToFuture:

/area control-plane
/kind bug
/platform openstack

Skip network mutation when shoot specs are equal. This may prevent deletion in some cases because the deletion confirmation annotation can not be added. See https://github.com/gardener/gardener/pull/7134 for such a case.

Fixes #

**Release note**:

```other operator
Skip network mutation when shoot specs are equal.
```


kon-angelo (approved):
/lgtm

PR #544, opened 2022-12-16
Validate worker zones for duplicates and existence in region

MartinWeindel:

/area control-plane
/kind enhancement
/platform openstack

Validate worker zones for duplicates and existence in region as specified in the cloud profile.

Fixes #

**Release note**:

```other user
Validate worker zones for duplicates and existence in region
```

### DIFF
--- a/pkg/apis/openstack/validation/shoot.go
+++ b/pkg/apis/openstack/validation/shoot.go
@@ -19,10 +19,12 @@ import (
 
 	"github.com/Masterminds/semver"
 	"github.com/gardener/gardener/pkg/apis/core"
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	validationutils "github.com/gardener/gardener/pkg/utils/validation"
 	"github.com/gardener/gardener/pkg/utils/version"
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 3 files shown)


pkg/apis/openstack/validation/shoot.go
```diff
@@ -114,8 +117,39 @@ func ValidateWorkers(workers []core.Worker, cloudProfileCfg *api.CloudProfileCon
...
+			if usedZones.Has(zone) {
+				allErrs = append(allErrs, field.Duplicate(fldPath.Index(i), zone))
+			}
+			usedZones.Insert(zone)
+			for _, z := range region.Zones {
+				if z.Name == zone {
+					continue outer
+				}
+			}
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i), zone, fmt.Sprintf("zone %s not existing in region %s", zone, region.Name)))
```
kon-angelo, line 144: I think that technically this check is already enforced on gardener level by the g/g admission checks.

MartinWeindel: Ah, ok. But duplicate zones are missing at the moment. I will take a look in the g/g admission checks. Thanks!

MartinWeindel:
enforced on gardener level by the g/g admission checks

kon-angelo:
@MartinWeindel the duplicate zones I think is indeed missing so the PR is still useful. Just the part that checks if the zones are part of a region is already been done.

PR #550, opened 2023-01-07
fix containerName of provider-openstack within the mcm Deployment

nschad:

/area scalability
/kind bug
/platform openstack

An incorrect `containerName` in the vpa object caused the container not beeing controlled by the vpa. In decent largish clusters this can cause the container to get `OOMKilled` as it only has 32Mi of memory.

Fixes #

~Helm diff might a bit large as my tar or whatever it is causing it produces a different output. But it does work 100%.~

**Release note**:

```bugfix operator
An incorrect container name in the machine-controller-manager Deployment caused the container not beeing controlled by the vpa.
```

### DIFF
--- a/charts/internal/machine-controller-manager/seed/templates/deployment.yaml
+++ b/charts/internal/machine-controller-manager/seed/templates/deployment.yaml
@@ -76,7 +76,7 @@ spec:
         - mountPath: /var/run/secrets/gardener.cloud/shoot/generic-kubeconfig
           name: kubeconfig
           readOnly: true
-      - name: provider-openstack
+      - name: machine-controller-manager-provider-openstack
         image: {{ index .Values.images "machine-controller-manager-provider-openstack" }}
         imagePullPolicy: IfNotPresent
         command:


gardener-robot-ci-1:
Thank you @nschad for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

MartinWeindel (dismissed):
/lgtm

ialidzhikov:
The verify step fails with:
```
>> make generate
make generate needs to be run:
 M example/controller-registration.yaml
```

@nschad , please revert the change in `example/controller-registration.yaml`. In this PR I don't see a change in the extension chart itself.

nschad:
> The verify step fails with:
> 
> ```
> >> make generate
> make generate needs to be run:
>  M example/controller-registration.yaml
> ```
> 
> @nschad , please revert the change in `example/controller-registration.yaml`. In this PR I don't see a change in the extension chart itself.

Done.

ialidzhikov (dismissed):
Thanks for the fix!

/lgtm

charts/internal/machine-controller-manager/seed/templates/vpa.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     updateMode: {{ .Values.vpa.updatePolicy.updateMode | quote }}
   resourcePolicy:
     containerPolicies:
-    - containerName: machine-controller-manager-provider-openstack
+    - containerName: provider-openstack
```
voelzmo, line 16: Sorry for being late to the party, I'm only seeing this now. That's a great catch, I suppose the reason why this happened is that this container has a different name in the openstack extension than in the other extensions. Therefore, can we fix it the other way round, for consistency reasons with the other extensions?
* aws extension container is named [`machine-controller-manager-provider-aws`](https://github.com/gardener/gardener-extension-provider-aws/blob/master/charts/internal/machine-controller-manager/seed/templates/deployment.yaml#L42)
* gcp extension container is named [`machine-controller-manager-provider-gcp`](https://github.com/gardener/gardener-extension-provider-gcp/blob/25adbe3828261d1ef1ffc9ad9b32ac90679157de/charts/internal/machine-controller-manager/seed/templates/deployment.yaml#L42)
* azure extension is named [`machine-controller-manager-provider-azure`](https://github.com/gardener/gardener-extension-provider-azure/blob/master/charts/internal/machine-controller-manager/seed/templates/deployment.yaml#L42)

Thanks!

voelzmo (changes requested):

charts/internal/machine-controller-manager/seed/templates/vpa.yaml
nschad: @voelzmo Sure. Do you want me to rename the container in the Deployment to `machine-controller-manager-provider-openstack` ?

nschad:
@voelzmo Done. Let me know if you need anything else :).

kon-angelo (approved):
/lgtm

PR #558, opened 2023-02-01
Automated cherry pick of #517 upstream release v1.31

DockToFuture:

/area networking
/kind enhancement
/platform openstack

This PR enables the overlay network as default for all new calico and cilium shoot cluster on openstack via the enable-overlay-as-default-for-calico and enable-overlay-as-default-for-cilium flag on the admission extension.
This is now the default behaviour but it can be changed by setting the enable-overlay-as-default-for-calico and enable-overlay-as-default-for-cilium flag to false on the openstack provider admission extension.
In case the native routing mode is used all traffic to the upstream dns server is now masqueraded for calico and cilium shoot clusters without overlay.
Additional tests were added to test the shoot mutation.

Fixes #

**Release note**:

```other operator
The overlay network is now globally enabled for all calico and cilium shoot clusters on openstack.
In case the overlay network is disabled all network traffic to the upstream dns server is now masqueraded for calico and cilium shoot clusters without overlay.
```


ScheererJ (approved):
/lgtm

kon-angelo (approved):
/lgtm

PR #563, opened 2023-02-03
Allow specifying CA cert for Keystone in Cloudprofile

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

1) By default the extension will not use `insecure` flag to talk to OpenStack apis.
2) A new flag in the cloudprofile is introduced to configure TF and clients to talk without cert validation ( = the current behavior).
3) Allows operators to specify a CA bundle in the Cloudprofile. 

Fixes #

**Release note**:

```breaking operator
The provider extension will not skip certificate verification by default. You can enable the old behavior by setting `KeyStoneForceInsecure` to true (per CloudProfile).
```
```feature operator
Allow specifying CA certificates for KeyStone api in Cloudprofile.
```


pkg/apis/openstack/types_cloudprofile.go
```diff
@@ -37,6 +37,10 @@ type CloudProfileConfig struct {
 	DHCPDomain *string
 	// KeyStoneURL is the URL for auth{n,z} in OpenStack (pointing to KeyStone).
 	KeyStoneURL string
+	// KeystoneCACert is the CA Bundle for the KeyStoneUR
```
MartinWeindel, line 40:
```suggestion
	// KeystoneCACert is the CA Bundle for the KeyStoneURL
```

pkg/apis/openstack/v1alpha1/types_cloudprofile.go
```diff
@@ -37,6 +37,12 @@ type CloudProfileConfig struct {
 	// KeyStoneURL is the URL for auth{n,z} in OpenStack (pointing to KeyStone).
 	// +optional
 	KeyStoneURL string `json:"keystoneURL,omitempty"`
+	// KeystoneCACert is the CA Bundle for the KeyStoneUR
```
MartinWeindel, line 40:
```suggestion
	// KeystoneCACert is the CA Bundle for the KeyStoneURL
```

MartinWeindel (dismissed):
/lgtm
Only found two nits.

MartinWeindel (approved):
/lgtm

pkg/openstack/client/client.go
```diff
@@ -46,10 +50,27 @@ func NewOpenstackClientFromCredentials(credentials *os.Credentials) (Factory, er
...
+
 	if opts.AuthInfo.ApplicationCredentialSecret != "" {
 		opts.AuthType = clientconfig.AuthV3ApplicationCredential
 	}
 
+	transport := &http.Transport{Proxy: http.ProxyFromEnvironment, TLSClientConfig: config}
+	opts.HTTPClient = &http.Client{
+		Transport: transport,
+	}
+
```
j2L4e, line 73: @kon-angelo It seems like settings opts.HTTPClient doesn't have any effect, since it isn't passed through to `authOpts`. Have you actually seen this work in action? 
Our feeling is that https://github.com/gardener/gardener-extension-provider-openstack/commit/72a3c33c79dd4bb5ae16fdbe6a9f243fe6ec8c99 might already be enough to tackle the problem.

@lotharbach @jensac

kon-angelo: I think that is reasonable. It is what gophercloud is also doing in it's utils package [ref](https://github.com/gophercloud/utils/blob/de873b9cf00d4b57e384f3f8e6a52dfcb290ecd8/openstack/clientconfig/requests.go#L826-L833). 

Thank you for pointing it that out @j2L4e. Can you open a PR for it ?

PR #568, opened 2023-02-08
Revert enabling overlay as default networking setup

kon-angelo:

/area control-plane
/kind bug
/platform openstack

This PR reverts https://github.com/gardener/gardener-extension-provider-openstack/pull/558 and https://github.com/gardener/gardener-extension-provider-openstack/pull/560. The current patch requires a new version of calico to be deployed which creates issues for current landscapes to consume.

Fixes #

**Release note**:

```other operator
The openstack validator will switch back to not using overlay as a default for shoots that do not specify any networking configuration.
```


MartinWeindel (approved):
/lgtm

PR #571, opened 2023-02-09
Disable overlay webhook completely

kon-angelo:

/area control-plane
/kind bug
/platform openstack

It modifies the webhook in-place so as to not make shoot mutations. We opted to change the webhook watched resources instead of deleting the mutating webhook, so as to avoid manual operations needed by operators to remove the mutating webhook and to make it easier to re-enable it in the future.

Fixes #

**Release note**:

```bugfix operator
Disable overlay webhook to prevent issues with overloading Neutron API with route requests.
```


MartinWeindel (approved):
/lgtm

MartinWeindel (approved):
/lgtm

PR #573, opened 2023-02-22
remove snat settings since they are no longer required

DockToFuture:

/area networking
/kind enhancement
/platform openstack

Remove the setting of enabling the snat traffic to the upstream dns sever because this is now the default option in the calico extension for non-overlay cluster and thus no longer required to be explicitly set in the openstack provider extension.
Furthermore do existing shoot clusters now keep also their set overlay field values when the `EnableOverlayAsDefault` field for calico or cilium is set. 

Fixes #

**Release note**:

```other operator
none
```

### DIFF
--- a/pkg/admission/mutator/shoot.go
+++ b/pkg/admission/mutator/shoot.go
@@ -91,6 +91,9 @@ func (s *shoot) Mutate(ctx context.Context, new, old client.Object) error {
 	switch shoot.Spec.Networking.Type {
 	case calico.ReleaseName:
 		overlay := &calicov1alpha1.Overlay{Enabled: false}
+		if EnableOverlayAsDefaultForCalico {
+			overlay = &calicov1alpha1.Overlay{Enabled: true}
+		}
 
 		networkConfig, err := s.decodeCalicoNetworkConfig(shoot.Spec.Networking.ProviderConfig)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/admission/mutator/shoot.go
```diff
@@ -112,15 +112,8 @@ func (s *shoot) Mutate(ctx context.Context, new, old client.Object) error {
 			}
 		}
 
-		if networkConfig.Overlay != nil && !networkConfig.Overlay.Enabled {
-			networkConfig.SnatToUpstreamDNS = &calicov1alpha1.SnatToUpstreamDNS{Enabled: true}
-		} else {
-			networkConfig.SnatToUpstreamDNS = &calicov1alpha1.SnatToUpstreamDNS{Enabled: false}
-		}
```
kon-angelo, line 119: Can't we summarize the change (line 92) like so:

```
overlay := &calicov1alpha1.Overlay{Enabled: false}
if EnableOverlayAsDefaultForCalico {
    overlay := &calicov1alpha1.Overlay{Enabled: true}
}
if oldShoot == nil && networkConfig.Overlay == nil {
    networkConfig.Overlay = overlay
}
...
```

and skip the post processing part ? Even more because the as it is right now it is not only enabling overlay "as default" but it applies it in all cases if I am not mistaken.

DockToFuture: It's not applying it in all cases, when the overlay field is not set for an old cluster, it will also not set it now. But we can change the order.

DockToFuture: With this change now as proposed we don't change it when the overlay field is already set for a cluster. We only change the overlay field value for new clusters or when the overlay field gets removed.

kon-angelo: @DockToFuture Can the overlay ever be removed ? Wouldn't it fall under: 

```
if oldShoot != nil && networkConfig.Overlay == nil
```

DockToFuture: Why would we want to remove this field once it was set? It could either be enabled or disabled.

kon-angelo (approved):
/lgtm

PR #576, opened 2023-02-28
Remove minAllowed.cpu from all components

voelzmo:

/area auto-scaling
/kind enhancement
/platform openstack

Remove all minAllowed.cpu values from VPA objects. Our tests didn't show harmful side effects of removing these and in many cases VPA did recommend values lower than the minAllowed. Therefore, we start saving resources if we switch to the VPA recommendation for CPU.

We don't do this for memory, as it is not compressible and would lead to more/faster OOMKills.

**Release note**:

```other operator
Removed `minAllowed.cpu` from all VPA objects
```


kon-angelo (approved):
/lgtm

PR #579, opened 2023-03-07
Add machine labels to worker pool

MartinWeindel:

/area control-plane
/kind enhancement
/platform openstack

A mechanism is needed to roll all machines of a worker pool on external infrastructure changes (e.g. vSMP MemoryOne chost). For this purpuse, machine labels are added to the `ProviderConfig` of the worker pool. They have a field `TriggerRollingOnUpdate` to mark them as relevant for machine rolling. This means that if such a label changes, all machines of the worker pool are rolled.

Fixes #

**Release note**:

```other operator
Add machine labels to the `ProviderConfig` of a worker pool to support rolling of its machines on label changes.
```


MartinWeindel:
Draft until clarified how/where to specify these special label names.

pkg/controller/worker/machines.go
```diff
@@ -223,16 +233,31 @@ func (w *workerDelegate) generateMachineConfig(ctx context.Context) error {
...
+		}
+	}
+	if len(pairs) > 0 {
+		sort.Strings(pairs)
+		additionalHashData = append(additionalHashData, pairs...)
+	}
+
+	// don't include complete provider config in Hash calculation
+	copy := pool
+	copy.ProviderConfig = nil
```
kon-angelo, line 258: i think the main problem we might have with this, is that current ServerGroup machines will be rolled out (because their providerConfig will change). This is a technical debt however and we might as well try to tackle it now than later. WDYT ?

pkg/apis/openstack/types_worker.go
```diff
@@ -66,6 +66,19 @@ type WorkerConfig struct {
...
+}
+
+// MachineLabel define key value pair to label machines.
+type MachineLabel struct {
+	// Name is the machine label key
+	Name string
+	// Value is the machine label value
+	Value string
+	// Roll controls if the machines should be rolled if the value changes
+	Roll bool
```
kon-angelo, line 81: I think we need something more descriptive e.g. `(force)Roll(out)OnUpdate`

pkg/controller/worker/machines.go
MartinWeindel: There is no perfect solution. With my last changes, it will always use the old hash code calculation as long as there are no machine labels. If a machine label is added (even if it is not setting `triggerRollingOnUpdate=true`), the new hash code calculation will be used and it causes a rolling of the machines as the hash changes. Later changes of not rolling machine labels will not change the hash.
In this way, no rolling happens for existing clusters, but we have the complication of two different hash code calculation. WDYT?

kon-angelo (approved):
/lgtm

PR #602, opened 2023-04-06
Improve worker hash calculation

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

Improve the hash calculation for machine-deployments. We should not include the providerConfig in the hash calculation and instead let the extension decide what is "important" enough to trigger a node rollout. This is a breaking change because changing the hash will trigger a node rollout. 

To ease the transition we include an annotation that shoot owners can use  to preserve the old hash calculation. This is done to allow shoot owners the freedom to make the node rollout at their discretion. Only worker pools using `server groups` should be affected by the change in the hash. However using the new annotation means **all** `machineLabels` will trigger a rollout. Users should complete the transition before using the `machineLabels` feature.

Fixes #

**Release note**:

```breaking user
Adapt worker node hash calculation. Shoots that are currently using the server group feature will be rolled out. You can prevent the rollout if the shoot is annotated with `openstack.provider.extensions.gardener.cloud/worker-preserve-hash`. 
```

### DIFF
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
+	"strings"
 
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/controller/worker"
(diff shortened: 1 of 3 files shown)


MartinWeindel (approved):
/lgtm

PR #613, opened 2023-04-26
add topology support for manila

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

+ Current storageclasses have now `volumeBindingMode: WaitForFirstConsumer`
+ Adds topology awareness support for Manila
+ Default deployment now is with TopologyAwareness
+ Fix an issue where the `nodeaz` was not computed correctly when enabled
+ Adds the following storageclasses:
  - A storage class without specifying the `availability` ( leaves it up to the Manila Scheduler to make decision about the location) 
  - Exclusive storageclasses per region (`*-constrain-*` variants). These storageclasses will make it so that both the provisioning and the scheduling of workloads using these volumes is exclusive to the target zone. 

Fixes #

**Release note**:

```other operator
Add topology awareness support for Manila
```

### DIFF
--- a/charts/internal/shoot-system-components/charts/csi-driver-manila/templates/daemonset-node.yaml
+++ b/charts/internal/shoot-system-components/charts/csi-driver-manila/templates/daemonset-node.yaml
@@ -42,22 +42,24 @@ spec:
               add: ["SYS_ADMIN"]
             allowPrivilegeEscalation: true
           image: {{ index .Values.images "csi-driver-manila" }}
-          args:
-            - /bin/manila-csi-plugin
-            - --nodeid=$(NODE_ID)
+          command:
+            - "/bin/sh"
+            - "-c"
+            - "/bin/manila-csi-plugin \
+              --nodeid=$(NODE_ID) \
             {{- if .Values.csimanila.runtimeConfig.enabled }}
-            - --runtime-config-file=/runtimeconfig/runtimeconfig.json
+              --runtime-config-file=/runtimeconfig/runtimeconfig.json \
             {{- end }}
             {{- if .Values.csimanila.topologyAwarenessEnabled }}
-            - --with-topology
-            - --nodeaz={{ .Values.csimanila.nodeAZ }}
+              --with-topology \
+              --nodeaz={{ .Values.csimanila.nodeAZ }}
             {{- end }}
-            - --endpoint=/csi/csi.sock
-            - --drivername=$(DRIVER_NAME)
-            - --share-protocol-selector=$(MANILA_SHARE_PROTO)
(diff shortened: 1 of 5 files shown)


kon-angelo:
cc: @kayrus

charts/internal/shoot-system-components/charts/csi-driver-manila/templates/daemonset-node.yaml
```diff
@@ -42,22 +42,24 @@ spec:
               add: ["SYS_ADMIN"]
             allowPrivilegeEscalation: true
           image: {{ index .Values.images "csi-driver-manila" }}
-          args:
-            - /bin/manila-csi-plugin
-            - --nodeid=$(NODE_ID)
+          command:
+            - "/bin/sh"
```
kayrus, line 46: What is the reason for switching to shell?

charts/internal/shoot-system-components/charts/csi-driver-manila/values.yaml
```diff
@@ -84,10 +84,10 @@ csimanila:
         }
       }
 
-  topologyAwarenessEnabled: false
+  topologyAwarenessEnabled: true
   # Availability zone for each node. topologyAwarenessEnabled must be set to true for this option to have any effect.
   # If your Kubernetes cluster runs atop of Nova and want to use Nova AZs as AZs for the nodes of the cluster, uncomment the line below:
-  #nodeAZ: "$(curl http://169.254.169.254/openstack/latest/meta_data.json | jq -r .availability_zone)"
+  nodeAZ: "$(curl http://169.254.169.254/openstack/latest/meta_data.json | jq -r .availability_zone)"
```
kayrus, line 90: probably this needs to be moved directly into the `csi-driver-manila/templates/daemonset-node.yaml`?

charts/internal/shoot-system-components/charts/csi-driver-manila/templates/daemonset-node.yaml
kon-angelo: Because without this the `--nodeaz` command was not being shell-substituted. Instead the whole command string was passed as the `nodeaz` value itself

kon-angelo: The upstream templates are also using the `/bin/sh` too.

kayrus: ah. right. the curl is executed here.

MartinWeindel (approved):
/lgtm

PR #623, opened 2023-05-11
specify fip subnet based on router

### LINKED ISSUE #615: subnetID not set when creating a floating IP for the bastion vm

/area ops-productivity
/kind bug
/platform openstack

**What happened**:

when creating a floating IP for the bastion vm the subnetID is not set on the CreateOpts [https://github.com/gardener/gardener-extension-provider-openstack/blob/e35f1b04e8a[…]ab94ff9613aef7e375/pkg/controller/bastion/actuator_reconcile.go](https://github.com/gardener/gardener-extension-provider-openstack/blob/e35f1b04e8a50c268306f0ab94ff9613aef7e375/pkg/controller/bastion/actuator_reconcile.go#L178-L181)
This could lead to wrong subnets being used for the floating IP of the bastion. 

**Environment**:

- Extension version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

This PR will make it so that the bastion controller binds FIPs from the same subnet as the router. We can potentially change the logic to use processing logic similar as with `FloatingPoolSubnetName`, but for simplicity this PR chooses the router subnet. The downside of that choice is that we depend on that particular subnet for available FIPs. 

Fixes #615 

**Release note**:

```other operator
The bastion with try to reserve Floating IPs from the router's external subnet
```

### DIFF
--- a/pkg/controller/bastion/actuator_reconcile.go
+++ b/pkg/controller/bastion/actuator_reconcile.go
@@ -179,20 +179,25 @@ func ensurePublicIPAddress(opt *Options, log logr.Logger, client openstackclient
 		return nil, errors.New("router must not be empty")
 	}
 
-	routers, err := client.GetRouterByID(infraStatus.Networks.Router.ID)
+	routerList, err := client.GetRouterByID(infraStatus.Networks.Router.ID)
 	if err != nil {
 		return nil, err
 	}
 
-	if len(routers) == 0 {
+	if len(routerList) == 0 {
 		return nil, fmt.Errorf("router with ID %s was not found", infraStatus.Networks.Router.ID)
 	}
 
-	var subnetID string
-	if router := routers[0]; len(router.GatewayInfo.ExternalFixedIPs) != 0 {
-		subnetID = router.GatewayInfo.ExternalFixedIPs[0].SubnetID
+	var (
+		subnetID string
+	)
+
+	router := routerList[0]
+	if len(router.GatewayInfo.ExternalFixedIPs) == 0 {
+		return nil, errors.New("no external fixed IPs detected on the router")
 	}
 
+	subnetID = router.GatewayInfo.ExternalFixedIPs[0].SubnetID
 	createOpts := floatingips.CreateOpts{
 		Description:       opt.BastionInstanceName,
 		FloatingNetworkID: infraStatus.Networks.FloatingPool.ID,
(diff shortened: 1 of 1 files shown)


pkg/controller/bastion/actuator_reconcile.go
```diff
@@ -175,9 +175,28 @@ func ensurePublicIPAddress(opt *Options, log logr.Logger, client openstackclient
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(routers) == 0 {
+		return nil, fmt.Errorf("router with ID %s was not found", infraStatus.Networks.Router.ID)
+	}
+
+	var subnetID string
+	if router := routers[0]; len(router.GatewayInfo.ExternalFixedIPs) != 0 {
```
petersutter, line 192: I'm not too deep in the topic. In which case is `ExternalFixedIPs` empty and is this considered an error?

kon-angelo: Practically it would mean that there is no FIP attached to the router which would be an error condition for the shoot. It would be an invalid shoot state so we can treat it as an error.

MartinWeindel (approved):
/lgtm

pkg/controller/bastion/actuator_reconcile.go
petersutter: will you make another commit to return an error in this case?

kon-angelo: @petersutter, yes I will update

pkg/controller/bastion/actuator_reconcile.go
```diff
@@ -175,9 +175,33 @@ func ensurePublicIPAddress(opt *Options, log logr.Logger, client openstackclient
...
+	var (
+		subnetID string
+	)
+
+	router := routerList[0]
+	if len(router.GatewayInfo.ExternalFixedIPs) == 0 {
+		return nil, errors.New("no external fixed IPs detected on the router")
+	}
+
+	subnetID = router.GatewayInfo.ExternalFixedIPs[0].SubnetID
```
petersutter, line 200:
```suggestion
	router := routerList[0]
	if len(router.GatewayInfo.ExternalFixedIPs) == 0 {
		return nil, errors.New("no external fixed IPs detected on the router")
	}

	subnetID := router.GatewayInfo.ExternalFixedIPs[0].SubnetID
```

petersutter (approved):
/lgtm

tedteng (approved):

MartinWeindel (approved):
/lgtm

PR #633, opened 2023-05-30
Allow changing share network section

MartinWeindel:

/area control-plane
/kind bug
/platform openstack

The admission webhook disallows changing network settings in the infrastructure config. But it also contains the `shareNetwork` section needed for CSI Manila support. This PR allows to enable or disable the share network for existing clusters.

Additionally it fixes the deletion of the CSI Manila controller deployment as a wrong name was specified ("csi-driver-manila-controller" instead of "csi-driver-controller-manila")

Fixes #

**Release note**:

```bugfix user
Allow changing share network section in `InfrastructureConfig` for existing cluster.
```

### DIFF
--- a/pkg/apis/openstack/validation/infrastructure.go
+++ b/pkg/apis/openstack/validation/infrastructure.go
@@ -83,7 +83,12 @@ func ValidateInfrastructureConfig(infra *api.InfrastructureConfig, nodesCIDR *st
 func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *api.InfrastructureConfig, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newConfig.Networks, oldConfig.Networks, fldPath.Child("networks"))...)
+	newNetworks := newConfig.Networks
+	oldNetworks := oldConfig.Networks
+	// share network changes are allowed, therefore ignore them on comparing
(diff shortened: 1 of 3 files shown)


pkg/apis/openstack/validation/infrastructure.go
```diff
@@ -83,7 +83,12 @@ func ValidateInfrastructureConfig(infra *api.InfrastructureConfig, nodesCIDR *st
 func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *api.InfrastructureConfig, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newConfig.Networks, oldConfig.Networks, fldPath.Child("networks"))...)
+	newNetworks := newConfig.Networks
+	oldNetworks := oldConfig.Networks
+	// share network changes are allowed, therefore ignore them on comparing
+	newNetworks.ShareNetwork = nil
+	oldNetworks.ShareNetwork = nil
```
kon-angelo, line 90: Is there any cause for concern for disabling a `shareNetwork` while there are volumes created ?

MartinWeindel: In this case the deletion of the `shareNetwork` will fail on the Openstack side. As a user you can fix it by removing the PVC/PV as long as the CSI manila driver is enabled.

kon-angelo (approved):
/lgtm

PR #636, opened 2023-06-02
fix incorrect manila values

kon-angelo:

/area control-plane
/kind regression
/platform openstack

Update the manila deployment with proper name and label selector (that were  changed in prior PR) , otherwise it would have prevented updated to existing deployments. As a result to enable logging for these containers, we also have to add the fluentbit configuration. 
Update VPA with the proper container names. 

Fixes #

**Release note**:

```other operator
NONE
```


MartinWeindel (approved):
/lgtm

PR #641, opened 2023-06-13
Improve cidr validation msg

dimityrmirchev:

/area quality
/kind enhancement
/platform openstack

Improves the message returned to the user when cidr validation fails for .networking.{nodes}.
Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/apis/openstack/validation/infrastructure.go
+++ b/pkg/apis/openstack/validation/infrastructure.go
@@ -36,9 +36,10 @@ func ValidateInfrastructureConfig(infra *api.InfrastructureConfig, nodesCIDR *st
 		allErrs = append(allErrs, field.Required(fldPath.Child("floatingPoolName"), "must provide the name of a floating pool"))
 	}
 
+	networkingPath := field.NewPath("networking")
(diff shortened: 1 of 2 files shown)


dimityrmirchev:
@kon-angelo I see that we validate only nodes cird here. Should we do the same for pods and services? 🤔

kon-angelo:
@dimityrmirchev what kind of checks are you suggesting ? The extensions typically only validate the CIDRs against infrastructure resources (so network and worker ranges). Gardener api-server is validating if the pod/service ranges are valid CIDRs and their correctness. The extension may still use these ranges e.g. for firewall rules.

kon-angelo:
@dimityrmirchev Ah, I see what you mean now. Yes we do check if the pods/services overlap with infrastructure ranges. It could be that the check is missing for openstack. Still, I think this matter shouldn't concern this PR.

dimityrmirchev:
> It could be that the check is missing for openstack.

Yes, exactly. 

> Still, I think this matter shouldn't concern this PR.

I agree.

kon-angelo (approved):
/lgtm

PR #649, opened 2023-07-06
Introduce persona-based docs

vlerenc:
We would like to introduce persona-based docs focussing on the already known release notes personas and therefore "cleanup" the structure somewhat:
- users (`/docs/usage`)
- operators (`/docs/operations`)
- developers (`/docs/development`)

cc @n-boshnakov

**Release note**:
None as ~30 PRs have/will be opened and the message would be the same everywhere (no relevant content changes either) while the only thing that actually changes in the end is the docs page that will have its release note.

/kind/cleanup


kon-angelo (approved):
/lgtm

PR #656, opened 2023-08-03
Clean leftover kubernetes loadbalancers

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

Clean leftover kubernetes Loadbalancers that block subnet deletion. In rare cases, the CCM will not delete loadbalancers which prevent subnet deletion and require manual action from operator usually. With this PR the LBs on the worker subnet should be automatically cleaned.

Please note, the Floating IP of the loadbalancers will not be deleted. This is, because users may want to preserve the Floating IP but without access to the k8s api server necessarily.

Fixes #

**Release note**:

```other operator
Provider-extension will attempt to delete all kubernetes loadbalancers that were not cleaned up by the CCM on infrastructure deletion.
```


MartinWeindel (approved):
/lgtm

PR #677, opened 2023-10-04
Remove dependency to specific calico and cilium versions.

axel7born:

/area networking
/kind technical-debt
/platform openstack

This PR removes the dependency to specific calico and cilium versions.
After each api change of a network extension, the new version needed to be vendored here and a new release has to be made in order to pass admission.
By parsing the network into a general map instead into an object of type `NetworkConfig` we can get rid of this dependency. 

Fixes #

**Release note**:

```noteworthy developer
Remove dependency to specific calico and cilium versions.
```


kon-angelo (approved):
/lgtm

pkg/admission/mutator/shoot.go
```diff
@@ -89,97 +93,57 @@ func (s *shoot) Mutate(_ context.Context, newObj, oldObj client.Object) error {
...
 
-			shoot.Spec.Networking.ProviderConfig = &runtime.RawExtension{
-				Object: networkConfig,
-			}
+		modifiedJSON, err := json.Marshal(networkConfig)
+		if err != nil {
+			return err
+		}
+		shoot.Spec.Networking.ProviderConfig = &runtime.RawExtension{
+			Raw: modifiedJSON,
```
kon-angelo, line 134: Do I remember correctly that the networking extension doesn't care at all about the `TypeMeta` information of the provided JSON ?

pkg/admission/mutator/shoot.go
```diff
@@ -44,6 +40,14 @@ type shoot struct {
 	decoder runtime.Decoder
 }
 
+const (
+	overlayKey         = "overlay"
+	enabledKey         = "enabled"
+	createPodRoutesKey = "createPodRoutes"
+	calicoReleaseName  = "calico"
+	ciliumReleaseName  = "cilium"
+)
```
kon-angelo, line 49: Would it look cleaner if it was a :

```

axel7born: Wouldn't we have then the same problem as before, that the config won't parse if there is a new field in the nertwork config?

kon-angelo (approved):
/lgtm

PR #686, opened 2023-10-30
readiness check

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

Adds a healthcheck to the snapshot-validation webhook server. Has the disadvantage that the logs will be spammed with messages like:

```
2023/10/30 18:30:38 http: TLS handshake error from 10.180.0.17:53028: EOF
```
This is to prevent issues like those introduced with external-snapshoter v6.3.0 where the server never started successfully and the bug took too long to detect.

Fixes #

**Release note**:

```other operator
Add tcp healthcheck to csi-snapshot-validation-webhook
```


MartinWeindel (approved):
/lgtm

PR #690, opened 2023-11-06
Support arm64

timebertt:

/area control-plane
/kind enhancement
/platform openstack

This PR adds support for arm64 shoot clusters.
This is realized by adding the `architecture` field to the machine image's region mappings (similar to https://github.com/gardener/gardener-extension-provider-aws/pull/565).

n/a

Operators could already declare machine image versions specific to a single architecture to workaround missing support for architecture-based mappings. E.g., add one image version `1.0_amd64` and one `1.0_arm64` and then map both individually in the `CloudProfileConfig`.
With this PR however, a missing `architecture` field in the machine image specification is assumed to express `amd64`.
I.e., if operators used this workaround for supporting non-amd64 machine images, this PR is breaking.
In that case, they should merge both versions into one and define architecture-based mappings as supported by this PR.

As arm64 flavors are still pretty seldom in OpenStack installation, it seems unlikely that somebody will run into issues with this breaking change.

**Release note**:

```feature operator
The `CloudProfileConfig` allows specifying machine image mappings by architecture. With this, the OpenStack extension can support `arm64` machine types.
```
```breaking operator
If operators specified machine image versions in their `CloudProfile` specific to the `arm64` architecture, they should migrate to an architecture-based mapping in the `CloudProfileConfig`.
```


gardener-robot-ci-1:
Thank you @timebertt for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

kon-angelo (approved):
/lgtm

MartinWeindel (approved):
/lgtm
I appreciate the comments at critical locations!

PR #698, opened 2023-12-11
Remove vendoring from project

afritzler:
/area dev-productivity
/kind enhancement
/platform openstack

***This is a WIP PR to illustrate how to remove vendoring from Gardener extension projects***

- Breaking webhook coding in Gardener has not been addressed here. 
- Fix `.ci` based vendor dependencies by copying the relevant CI files into a `hack` subfolder
- Fix references in `Makefile` by using `go list` to determine where to find the Gardener `hack` scripts
- Fix `go generate` markers to use the correct script location by injecting the `GARDENER_HACK_DIR` location via `Makefile`
- Use `go mod download` in `Dockerfile` to prefetch deps
- Use `etcd-druid-api` instead of `etcd-druid`

/cc @kon-angelo @MartinWeindel 

**Release note**:
```other developer
Remove vendoring from project
```


.ci/component_descriptor
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-"$(dirname $0)"/../vendor/github.com/gardener/gardener/hack/.ci/component_descriptor "$(dirname $0)"/..
+"$(dirname $0)"/hack/gardener_component_descriptor "$(dirname $0)"/..
```
MartinWeindel, line 3: fix filename
```suggestion
"$(dirname $0)"/hack/component_descriptor "$(dirname $0)"/..
```

Makefile
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+GARDENER_HACK_DIR           := $(shell go list -m -f "{{.Dir}}" github.com/gardener/gardener)/hack
```
MartinWeindel, line 15: ensure gardener module directtory available
```suggestion
ENSURE_GARDENER_MOD         := $(shell go get github.com/gardener/gardener@$$(go list -m -f "{{.Version}}" github.com/gardener/gardener))
GARDENER_HACK_DIR           := $(shell go list -m -f "{{.Dir}}" github.com/gardener/gardener)/hack
```

MartinWeindel (changes requested):
suggestions to fix build errorrs

.ci/hack/component_descriptor
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env bash
```
vpnachev: Are those hack scripts now copied and they have to be maintained in each repository using them or they are somehow synced from a single source repository?

afritzler: They are copied via `make tidy`. Essentially that happened in the past via `make revendor` and stored in the `vendor` folder. Unfortunately in the Councourse CI environment we can't rely on a golang environment to be present so we can't reference them dynamically via `go mod`/`go list`.

vpnachev: That is very nice, thanks for the clarification!

afritzler:
/ref https://github.com/gardener/etcd-druid/issues/734

Currently `etcd-druid` is on a super old Gardener version (1.76). This leads to some weird behaviour when updating the latest Gardener commit version. Adding a ref here to keep track of this issue.

timuthy:
As discussed, replaced by #704
/close

Issue #664, opened 2023-09-06
Support for enabling share networks is missing in the Flow-based infrastructure reconciliation without Terraformer


MartinWeindel:
**How to categorize this issue?**

/area control-plane
/kind enhancement
/platform openstack

**What would you like to be added**:
Support for enabling share networks is missing in the Flow-based infrastructure reconciliation without Terraformer

**Why is this needed**:

PR #715, opened 2024-02-02
Add support for shared network in the flow reconciler

kon-angelo:

/area control-plane
/kind technical-debt
/platform openstack

Fixes https://github.com/gardener/gardener-extension-provider-openstack/issues/664

**Release note**:

```other operator
Add support for share networks in the flow reconciler.
```


MartinWeindel (changes requested):
Implementation looks good, but it creates the "ShareNetwork" unconditionally and is ignoring the  field `spec.provider.infrastructureConfig.networks.shareNetwork.enabled` (compare with https://github.com/gardener/gardener-extension-provider-openstack/blob/master/pkg/internal/infrastructure/terraform.go#L133).
Was it a conscious decision or did you just missed it?

Of course, if the flag is considered, the share network may also be needed to be deleted during reconciliation if the flag is changed from `true` to `false`.

kon-angelo:
@MartinWeindel ptal :)

MartinWeindel (approved):
/lgtm

PR #731, opened 2024-03-19
Fix subtarget names of docker-images target

j2L4e:
Neither `docker-images-provider` nor `docker-images-admission` exist. Use `image` instead of `images` in both to make `make docker-images` work again.


kon-angelo (approved):
/lgtm

PR #733, opened 2024-03-19
Set lb-method="SOURCE_IP_PORT" when using "ovn" loadbalancer provider

j2L4e:
/area networking
/kind enhancement
/platform openstack

The extension doesn't allow setting a loadbalancer's lb-method attribute. The default for lb-method is `ROUND_ROBIN`, but the ovn loadbalancer provider only supports `SOURCE_IP_PORT`. This renders ovn unusable. Since `SOURCE_IP_PORT` is to be used when and only when the provider is ovn, and ovn doesn't work without it, it seems reasonable to just hard-code the value.

> `lb-method` The load balancing algorithm used to create the load balancer pool. [...] If lb-provider is set to "ovn" the value must be set to SOURCE_IP_PORT.
(https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#load-balancer)

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/internal/cloud-provider-config/templates/cloud-provider-config-loadbalancer.tpl
+++ b/charts/internal/cloud-provider-config/templates/cloud-provider-config-loadbalancer.tpl
@@ -6,6 +6,9 @@ monitor-timeout="{{ .Values.monitorTimeout }}"
 monitor-max-retries={{ .Values.monitorMaxRetries }}
 lb-version="v2"
 lb-provider="{{ .Values.lbProvider }}"
+{{- if eq .Values.lbProvider "ovn" }}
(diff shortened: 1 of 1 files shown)


gardener-robot-ci-2:
Thank you @j2L4e for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

j2L4e:
cc @Lappihuan

kon-angelo (approved):
/lgtm

PR #750, opened 2024-04-15
Set responsible label in component descriptor

AndreasBurger:
Adds component-responsible label to the component descriptor. This way, the responsibles will also be recognized for findings on images that are referenced, not only for those that are built.
Also, simplify by referring to the maintainer team.


hebelsan (approved):
/lgtm

PR #752, opened 2024-04-16
Add cluster-autoscaler options to machine deployment

aaronfern:

/area auto-scaling
/kind enhancement
/platform openstack

This PR adds cluster-autoscaler options for worker groups from workers to machine deployments. These options allow users to specify some cluster-autoscaler specific for worker groups

Fixes partially gardener/autoscaler#240

- `Autoscaler` will read these values from annotations in the `machineDeployment` (ref: https://github.com/gardener/autoscaler/pull/257)
- These values are specified in the shoot via `spec.provider.workers.clusterAutoscaler` and are added to the worker (ref: https://github.com/gardener/gardener/pull/9245)

**Release note**:

```other operator
NodeGroupAutoscalingOptions can now be specified per worker group via the worker through the field `worker.spec.pools.clusterAutoscaler`
```


kon-angelo (approved):
/lgtm

PR #779, opened 2024-05-28
update flow reconciler

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

Fixes #

**Release note**:

```other operator
QoL improvements to the infrastructure reconciler
```
```other operator
Add migration test for the infrastructure reconciler
```


pkg/controller/infrastructure/infraflow/shared/basic_context.go
```diff
@@ -142,21 +144,46 @@ func (c *BasicFlowContext) AddTask(g *flow.Graph, name string, fn flow.TaskFn, o
 	return g.Add(task)
 }
 
+// wrapTaskFn sets up the task function fn. It wraps it with the hooks
```
AndreasBurger, line 147:
```suggestion
// wrapTaskFn sets up the task function. It wraps it with the hooks
```

pkg/controller/infrastructure/infraflow/shared/whiteboard.go
```diff
@@ -258,10 +274,25 @@ func IsValidValue(value string) bool {
...
-		return ""
-	}
-	return value
+// HasObject returns true if the object with the given key exists.
+func (w *whiteboard) HasObject(key string) bool {
+	w.Lock()
+	defer w.Unlock()
+	_, ok := w.objects[key]
+	return ok
+}
```
AndreasBurger, line 283: I don't object to doing it this way, but checking for the existence of a key should be possible without locking the datastructure. We don't actually gain anything by locking here.

pkg/controller/infrastructure/terraform_reconciler.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		return err
+	}
+
+	return infrastructure.PatchProviderStatusAndState(ctx, t.client, infra, status, state)
+}
+
+// Delete deletes the infrastructure using Terraformer.
+func (t *TerraformReconciler) Delete(ctx context.Context, infra *extensionsv1alpha1.Infrastructure, c *extensions.Cluster) error {
+	return util.DetermineError(t.delete(ctx, infra, c), helper.KnownCodes)
```
AndreasBurger, line 112: Some codepaths in `delete` already call `DetermineError`. If this needs to be done always, we might _just_ do it here. Otherwise, if this does not need to be done in _some_ cases I would remove this call to `DetermineError`

pkg/internal/infrastructure/infrastucture.go
```diff
@@ -142,3 +146,42 @@ func WorkersCIDR(config *openstack.InfrastructureConfig) string {
...
+	state *runtime.RawExtension,
+) error {
+
+	// infraObjectKey := client.ObjectKey{
+	// 	Namespace: infra.Namespace,
+	// 	Name:      infra.Name,
+	// }
+	// if err := runtimeClient.Get(ctx, infraObjectKey, infra); err != nil {
+	// 	return err
+	// }
```
AndreasBurger, line 165:
```suggestion
```

pkg/internal/infrastructure/terraform.go
```diff
@@ -52,7 +52,7 @@ const (
 	MaxApiCallRetries = "10"
 )
 
-// StatusTypeMeta is the TypeMeta of the GCP InfrastructureStatus
+// StatusTypeMeta is the TypeMeta of the InfrastructureStatus
 var StatusTypeMeta = metav1.TypeMeta{
 	APIVersion: apiv1alpha1.SchemeGroupVersion.String(),
 	Kind:       "InfrastructureStatus",
```
AndreasBurger, line 58: I keep seeing the kind hardcoded in the code - is there no const available somewhere?
Just wondering, I don't see us actually changing the Kind anytime soon.

AndreasBurger (dismissed):
mostly lgtm

charts/internal/seed-controlplane/charts/csi-driver-manila-controller/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
-description: Helm chart for csi-driver-manila-controller (shared file system)
+description: Helm chart for csi-driver-manila-controller (shared2 file system)
```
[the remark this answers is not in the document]
kon-angelo: Unfortunate refactoring 😅

pkg/controller/infrastructure/infraflow/shared/whiteboard.go
kon-angelo: This is copied over from the other shared packages. Also not used in this provider at all. But regarding the access, I think that if you use mutexes then you should lock everything, or you can implement it with RWLocks. We either way only lock access to the map and not the objects.

pkg/controller/infrastructure/infraflow/shared/basic_context.go
kon-angelo: `fn` is the name of the argument

pkg/controller/infrastructure/infraflow/shared/whiteboard.go
AndreasBurger: > This is copied over from the other shared packages.
Yeah, I figured. Maybe we can find the time to go over the whiteboard for the second round of refactoring, but it's not actually causing errors so it's fine either way.
>  But regarding the access, I think that if you use mutexes then you should lock everything […]
Read-only access to a shared datastructures in general should not require locks. Is there a case where _not_ locking could  cause in a race condition? I can't think of one (that is not also possible with locking, that is). I'll have to read up on go-maps at some point.

AndreasBurger (approved):

hebelsan (dismissed):
lgtm - just some minor comments

hebelsan (approved):

pkg/controller/infrastructure/infraflow/utils.go
```diff
@@ -34,6 +33,12 @@ func findExisting[T any](id *string, name string,
 	if len(found) == 0 {
 		return nil, nil
 	}
+
+	// TODO: check if this makes sense
```
hebelsan, line 37: I guess this makes sense

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -27,344 +29,347 @@ const (
...
 }
 
-func (c *FlowContext) ensureConfiguredRouter(_ context.Context) error {
-	router, err := c.access.GetRouterByID(c.config.Networks.Router.ID)
+func (fctx *FlowContext) ensureConfiguredRouter(_ context.Context) error {
+	router, err := fctx.access.GetRouterByID(fctx.config.Networks.Router.ID)
 	if err != nil {
-		c.state.Set(IdentifierRouter, "")
+		// todo: Remove ?
+		fctx.state.Set(IdentifierRouter, "")
```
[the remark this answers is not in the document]
kon-angelo: Well, if he deletes the router then next conditional (`	if router == nil`) should be true (and in which case we should delete from our state - as you said). 

In this LoC, it's more of a question about issues with e.g. credentials not being renewed or issues with the API (say a 500 internal error), in which case we would remove the ID we had saved.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -27,344 +29,347 @@ const (
...
 		log.Info("looking up floating pool subnets...")
-		desired.ExternalSubnetIDs, err = c.access.LookupFloatingPoolSubnetIDs(externalNetworkID, *floatingPoolSubnetName)
+		desired.ExternalSubnetIDs, err = fctx.access.LookupFloatingPoolSubnetIDs(externalNetworkID, *floatingPoolSubnetName)
 		if err != nil {
 			return err
 		}
 	}
 	log.Info("creating...")
-	created, err := c.access.CreateRouter(desired)
+	// TODO: add tags to created resources
```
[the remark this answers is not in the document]
kon-angelo: @hebelsan sure there are https://github.com/gophercloud/gophercloud/blob/3bb883c4ebe42c7eb9382406e4b3f9e1cc3743ca/openstack/networking/v2/extensions/layer3/routers/results.go#L79 😉, but you need "special" APIs to create them (aka it's not done by the default `createOpts`). But it's a bit out of scope for this PR (hence, only the todo)

pkg/controller/infrastructure/infraflow/reconcile.go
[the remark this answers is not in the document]
kon-angelo: Nonetheless, I agree that we should keep it 👍

pkg/controller/infrastructure/infraflow/utils.go
kon-angelo: I guess so. The `findExisting` is geared towards finding a single result. I think it's better to return an error if we do find multiple instead of arbitrarily picking one of them (past behavior).

pkg/controller/infrastructure/infraflow/reconcile.go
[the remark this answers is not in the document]
hebelsan: I also believe in the cc dashboard router tags are not (yet) visible.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -423,49 +428,50 @@ func (c *FlowContext) ensureSecGroupRules(ctx context.Context) error {
...
 		log.Info("replacing SSH key pair")
-		if err := c.compute.DeleteKeyPair(c.namespace); err != nil {
+		if err := fctx.compute.DeleteKeyPair(fctx.defaultSSHKeypairName()); err != nil {
 			return err
 		}
 		keyPair = nil
+		fctx.state.Set(NameKeyPair, "")
 	}
 
 	if keyPair == nil {
```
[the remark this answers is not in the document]
kon-angelo: Correct, but it should be correct. The func will terminate early on:
```
		if keyPair.PublicKey == string(fctx.infra.Spec.SSHPublicKey) {
			fctx.state.Set(NameKeyPair, keyPair.Name)
			return nil
		}
```
So we just check the public key and if it matches then there is nothing else to do. In all other cases (either we don't find it or it doesn't match) we delete the keypair which is why we always set it to nil at that point.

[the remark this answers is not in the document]
hebelsan: I agree the flow is correct, just the if condition in line 451 `if keyPair == nil` can be omitted.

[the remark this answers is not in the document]
kon-angelo: Ah, I was being oblivious to it. Nice catch !

AndreasBurger (approved):

PR #780, opened 2024-05-28
Add support for deployments in existing subnet

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

/invite @Kumm-Kai 

Fixes #

**Release note**:

```other operator
Add support for creating shoots in existing subnets 
```


kon-angelo:
bastion test is flaky but the infra tests seem stable now. @AndreasBurger @hebelsan  feel free to ptal

pkg/apis/openstack/validation/controlplane.go
```diff
@@ -138,13 +138,18 @@ func validateLoadBalancerClassesConstraints(floatingPools []api.FloatingPool, sh
...
+	// for an existing subnet we do not need to validate the storage since we can only do that in runtime.
+	if networks.SubnetID != nil {
+		return allErrs
 	}
-	return allErrs
+	if networks.ShareNetwork != nil && networks.ShareNetwork.Enabled {
+		return allErrs
+
+	}
+	return append(allErrs, field.Invalid(fldPath.Child("csiManila", "enabled"), storage.CSIManila.Enabled, "share network must be created if CSI manila driver is enabled"))
```
AndreasBurger, line 154: Doing it like this will not necessarily detect invalid configurations, e.g. `networks.ShareNetwork.Enabled` being set but `ShareNetwork` not being given while `networks.SubnetID` is given.
I suspect that won't cause issues in the actual code. Still, not sure if short-cutting in validation is the way to go, I'd prefer it to be exhaustive and explicit.

pkg/apis/openstack/validation/infrastructure.go
```diff
@@ -33,7 +33,7 @@ func ValidateInfrastructureConfig(infra *api.InfrastructureConfig, nodesCIDR *st
 	}
 
 	networksPath := fldPath.Child("networks")
-	if len(infra.Networks.Worker) == 0 && len(infra.Networks.Workers) == 0 {
+	if len(infra.Networks.Worker) == 0 && len(infra.Networks.Workers) == 0 && infra.Networks.SubnetID == nil {
```
AndreasBurger, line 36:
```suggestion
	if len(infra.Networks.Worker) == 0 && infra.Networks.SubnetID == nil {
```
redundant

pkg/controller/infrastructure/configvalidator.go
```diff
@@ -70,6 +74,18 @@ func (c *configValidator) Validate(ctx context.Context, infra *extensionsv1alpha
 	// Validate infrastructure config
 	logger.Info("Validating infrastructure configuration")
 	allErrs = append(allErrs, c.validateFloatingPoolName(ctx, networkingClient, config.FloatingPoolName, field.NewPath("floatingPoolName"))...)
+	if config.Networks.ID != nil {
+		allErrs = append(allErrs, c.validateNetwork(ctx, networkingClient, *config.Networks.ID, field.NewPath("networks.id"))...)
+	}
+	if config.Networks.SubnetID != nil {
+		allErrs = append(allErrs, c.validateSubnet(ctx, networkingClient, *config.Networks.SubnetID, *config.Networks.ID, field.NewPath("networks.subnetId"))...)
+	}
+	if config.Networks.Router != nil && config.Networks.Router.ID != "" {
```
hebelsan, line 83: Imo it's unnecessary to check for `config.Networks.Router.ID != ""` because why should you be able to set the router config without an ID in the first place.

pkg/controller/infrastructure/configvalidator.go
```diff
@@ -91,3 +107,76 @@ func (c *configValidator) validateFloatingPoolName(ctx context.Context, networki
...
+
+	return allErrs
+}
+
+func (c *configValidator) validateRouter(_ context.Context, networkingClient openstackclient.Networking, routerID string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	routers, err := networkingClient.ListRouters(routers.ListOpts{ID: routerID})
+	if err != nil {
+		allErrs = append(allErrs, field.InternalError(fldPath, fmt.Errorf("could not get subnet: %w", err)))
```
hebelsan, line 153: you probably mean "could not get routers"

pkg/controller/infrastructure/configvalidator.go
```diff
@@ -91,3 +107,76 @@ func (c *configValidator) validateFloatingPoolName(ctx context.Context, networki
...
+	}
+
+	if port == nil {
+		allErrs = append(allErrs, field.Required(fldPath, fmt.Sprintf("could not get router interface for router [%s] and subnet [%s]", routerID, subnetID)))
+	}
+
+	return allErrs
+}
+
+func isValidString(s *string) bool {
```
AndreasBurger, line 180:
```suggestion
func isNotEmpty(s *string) bool {
```

pkg/controller/infrastructure/infraflow/delete.go
```diff
@@ -72,14 +79,17 @@ func (fctx *FlowContext) buildDeleteGraph() *flow.Graph {
...
+	)
 	deleteRouterInterface := fctx.AddTask(g, "delete router interface",
 		fctx.deleteRouterInterface,
+		shared.DoIf(needToDeleteSubnet || needToDeleteRouter),
 		shared.Timeout(defaultTimeout), shared.Dependencies(recoverRouterID, recoverSubnetID, k8sRoutes))
 	// subnet deletion only needed if network is given by spec
 	_ = fctx.AddTask(g, "delete subnet",
 		fctx.deleteSubnet,
-		shared.DoIf(!needToDeleteNetwork), shared.Timeout(defaultTimeout), shared.Dependencies(deleteRouterInterface, k8sLoadBalancers))
+		shared.DoIf(!needToDeleteNetwork && needToDeleteSubnet), shared.Timeout(defaultTimeout), shared.Dependencies(deleteRouterInterface, k8sLoadBalancers))
```
AndreasBurger, line 92:
```suggestion
                // Delete the subnet only if the network is not being deleted already since it will be deleted anyways in that case.
		shared.DoIf(needToDeleteSubnet && !needToDeleteNetwork ), shared.Timeout(defaultTimeout), shared.Dependencies(deleteRouterInterface, k8sLoadBalancers))
```

pkg/controller/infrastructure/infraflow/delete.go
```diff
@@ -36,6 +36,8 @@ func (fctx *FlowContext) buildDeleteGraph() *flow.Graph {
 
 	needToDeleteNetwork := fctx.config.Networks.ID == nil
 	needToDeleteRouter := fctx.config.Networks.Router == nil
+	// skip deletion of the subnet if we need to delete the network (it will be deleted anyway), or if the subnet is user provided.
```
AndreasBurger, line 39:
```suggestion
	// skip deletion of the subnet if it is user provided.
```
I've moved the other part of the comment closer to the relevant line (see below)

hebelsan (changes requested):

pkg/internal/infrastructure/templates/main.tpl.tf
```diff
@@ -80,11 +81,23 @@ resource "openstack_networking_subnet_v2" "cluster" {
   dns_nameservers = []
   {{- end }}
 }
+{{ else -}}
+data "openstack_networking_subnet_v2" "cluster" {
+  subnet_id = "{{ .networks.subnet }}"
+}
+{{- end }}
 
+// This check is needed since the "Use existing subnet" feature
```
AndreasBurger, line 90:
```suggestion
// This check is needed for the "Use existing subnet" feature
```

AndreasBurger (changes requested):
mostly lgtm, one or two nits

pkg/controller/infrastructure/configvalidator.go
```diff
@@ -70,6 +74,18 @@ func (c *configValidator) Validate(ctx context.Context, infra *extensionsv1alpha
 	// Validate infrastructure config
 	logger.Info("Validating infrastructure configuration")
 	allErrs = append(allErrs, c.validateFloatingPoolName(ctx, networkingClient, config.FloatingPoolName, field.NewPath("floatingPoolName"))...)
+	if config.Networks.ID != nil {
+		allErrs = append(allErrs, c.validateNetwork(ctx, networkingClient, *config.Networks.ID, field.NewPath("networks.id"))...)
+	}
+	if config.Networks.SubnetID != nil {
+		allErrs = append(allErrs, c.validateSubnet(ctx, networkingClient, *config.Networks.SubnetID, *config.Networks.ID, field.NewPath("networks.subnetId"))...)
```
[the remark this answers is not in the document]
kon-angelo: Sure. It's one of these things that should be mandated by the validation at the admission controller, but it's a fair point

pkg/controller/infrastructure/configvalidator.go
kon-angelo: This is a bit contradicting with the comment above. The controller that validates that you have a subnet.id iff a network.id exists is the same that performs the validation about the empty router.id.

hebelsan: fair point, I'm fine with leaving it as is.

hebelsan (dismissed):
/lgtm

charts/internal/shoot-system-components/charts/csi-driver-manila/templates/storageclass.yaml
```diff
@@ -15,7 +15,9 @@ mountOptions:
 parameters:
   type: default
   shareNetworkID: {{ $.Values.openstack.shareNetworkID }}
+  {{- if $.Values.openstack.shareClient}}
   nfs-shareClient: {{ required "openstack.shareClient needs to be set" $.Values.openstack.shareClient }}
```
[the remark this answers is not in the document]
kon-angelo: Doesn't this depend a little bit on how the "other side" and if it can parse the empty values ?

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -995,7 +990,6 @@ func (vp *valuesProvider) addCSIManilaValues(
 	values["openstack"] = map[string]interface{}{
 		"availabilityZones":           vp.getAllWorkerPoolsZones(cluster),
 		"shareNetworkID":              shareNetworkID,
-		"shareClient":                 infrastructure.WorkersCIDR(infraConfig),
```
hebelsan, line 998: How can the "shareClient" be set if we remove it from here?

hebelsan (changes requested):
We should probably add documentation for bringing your own subnetID.

pkg/controller/controlplane/valuesprovider.go
kon-angelo: We don't set it basically. The shareClient constricts who can access the share, for which at the moment we have no strict need to do, it even restrict customer use cases. The template doesn't necessarily  need to change, we just choose not to configure this.

PR #794, opened 2024-06-21
Include absent router in infra-dependency errs

AndreasBurger:
/area disaster-recovery
/kind enhancement
/platform openstack

With this PR, the router being absent will be considered an `ERR_INFRA_DEPENDENCIES` error, enabling force deletion on the shoot.

I've done some tests on local shoots and this seems to only occur for shoots created using Terraform. For shoots created via our flow, deletion was able to be completed successfully even when the router was removed.

**Release note**:

```other user
When a missing router is being encountered during reconciliation/deletion it will now be classified as `ERR_INFRA_DEPENDENCIES`, enabling force deletion of the shoot via annotation.
```


hebelsan (approved):
/lgtm

PR #825, opened 2024-07-31
[Release 1.41] fix an issue where shoots would be migrated to flow during cp-migration

AndreasBurger:

/area control-plane
/kind bug
/platform openstack

Fixes #

**Release note**:

```other operator
Fix a bug that would migrate shoots away from terraform during control-plane migration. This would occur only if the seed was marked to use the flow reconciler for new shoots.
```


kon-angelo (approved):
/lgtm

PR #829, opened 2024-08-05
Update dependencies in images.yaml

AndreasBurger:


kon-angelo:
close after https://github.com/gardener/gardener-extension-provider-openstack/pull/865

PR #840, opened 2024-08-28
Return err instead of nil

hebelsan:

/area robustness
/kind bug
/platform openstack

Fix case where nil instead of the actual error was returned

Fixes #

**Release note**:

```other bugfix
Fix case where nil instead of the actual error was returned
```


kon-angelo (approved):
/lgtm

PR #870, opened 2024-10-04
Improve confusing error message

ialidzhikov:

/area ops-productivity
/kind enhancement
/platform openstack

The error message 
```
[...]: failed to create openstack seedClient: Authentication failed
```

is confusing. There openstack client is actually an openstack client against the Shoot's infrastructure. Maybe we wanted to express a failure creating an openstack client using a seedClient (which is a K8s client)? IMO there is no need to provide such internal details.

N/A

N/A

**Release note**:

```other operator
NONE
```


kon-angelo (approved):
/lgtm

kon-angelo:
Thank you, more like a refactoring error

PR #864, opened 2024-09-26
Update cloud-provider-openstack images and remove credential use from nodeserver

kon-angelo:
**How to categorize this PR?**

/area control-plane
/kind enhancement
/platform openstack

**What this PR does / why we need it**:

- Update cloudprovider images (including CSI driver images)
- Remove use of credentials from the nodeserver daemonset
- Remove deprecated v1.25 images

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator
Update cloud-provider-openstack images and remove credential use from nodeserver
```

PR #875, opened 2024-10-11
Fix k8s <=v1.27 deployment of csi-driver-node

kon-angelo:

/area control-plane
/kind bug
/platform openstack

Fix an issue after https://github.com/gardener/gardener-extension-provider-openstack/pull/864 that prevented the successful deployment of the csi-controller and csi-node on k8s <=1.27

Fixes #

**Release note**:

```other operator
Fix a bug that prevented the Cinder CSI from working on shoots with kubernetes version less than v1.28.
```


hebelsan (approved):
/lgtm

Issue #890, opened 2024-11-01
If shoot creation fails on `ensureNetwork`, then cluster cannot be deleted


RadaBDimitrova:
**How to categorize this issue?**

/area networking
/kind bug
/platform openstack

**What happened**:

On shoot creation when we reach "ensure network" here:
https://github.com/gardener/gardener-extension-provider-openstack/blob/c14529a1d7c69e071a077ae1c88d67a24a75eab9/pkg/controller/infrastructure/infraflow/reconcile.go#L58-L62

The reconciliation flow graph can be severed in case of a failure in the flow task, for example here:
https://github.com/gardener/gardener-extension-provider-openstack/blob/c14529a1d7c69e071a077ae1c88d67a24a75eab9/pkg/controller/infrastructure/infraflow/reconcile.go#L240-L246

After execution of the task the following state was observed, so the network wasn't set:
```
state:
    apiVersion: openstack.provider.extensions.gardener.cloud/v1alpha1
    data:
      FloatingNetwork: ...
      FloatingNetworkName: ...
    kind: InfrastructureState
```
Shoot gets stuck in creation state and after triggering shoot deletion the following snippet fails the "recover subnet ID" task with "network not found":
https://github.com/gardener/gardener-extension-provider-openstack/blob/c14529a1d7c69e071a077ae1c88d67a24a75eab9/pkg/controller/infrastructure/infraflow/reconcile.go#L311-L318

**What you expected to happen**:
Shoot deletion to succeed after reconciliation failed.

**How to reproduce it (as minimally and precisely as possible)**:
N/A

kon-angelo:
@RadaBDimitrova I agree that the issue does occur but the `getNetworkID` also checks the infrastructure for the network. So the issue I believe is not that the network was not set in the state, but rather that it was never created.

PR #898, opened 2024-11-06
fix deletion when network doesn't exist

kon-angelo:

/area control-plane
/kind bug
/platform openstack

Fixes #https://github.com/gardener/gardener-extension-provider-openstack/issues/890

**Release note**:

```other operator
Fix an issue where the deletion with the flow reconciler would fail if the network was already deleted.
```

### DIFF
--- a/pkg/controller/infrastructure/infraflow/delete.go
+++ b/pkg/controller/infrastructure/infraflow/delete.go
@@ -103,48 +103,47 @@ func (fctx *FlowContext) buildDeleteGraph() *flow.Graph {
 }
 
 func (fctx *FlowContext) deleteRouter(ctx context.Context) error {
-	log := shared.LogFromContext(ctx)
-	current, err := fctx.findExistingRouter()
-	if err != nil {
+	routerID := fctx.state.Get(IdentifierRouter)
+	if routerID == nil {
+		return nil
+	}
+
+	shared.LogFromContext(ctx).Info("deleting...", "router", *routerID)
+	if err := fctx.networking.DeleteRouter(*routerID); err != nil {
 		return err
 	}
-	if current != nil {
-		log.Info("deleting...", "router", current.ID)
-		if err := fctx.networking.DeleteRouter(current.ID); err != nil {
-			return err
-		}
-	}
+
+	fctx.state.Set(IdentifierRouter, "")
 	return nil
 }
 
 func (fctx *FlowContext) deleteNetwork(ctx context.Context) error {
-	log := shared.LogFromContext(ctx)
-	current, err := fctx.findExistingNetwork()
-	if err != nil {
+	networkID := fctx.state.Get(IdentifierNetwork)
+	if networkID == nil {
+		return nil
+	}
+
+	shared.LogFromContext(ctx).Info("deleting...", "network", *networkID)
+	if err := fctx.networking.DeleteNetwork(*networkID); err != nil {
 		return err
 	}
-	if current != nil {
-		log.Info("deleting...", "network", current.ID)
-		if err := fctx.networking.DeleteNetwork(current.ID); err != nil {
-			return err
-		}
-	}
+
 	fctx.state.Set(NameNetwork, "")
+	fctx.state.Set(IdentifierNetwork, "")
 	return nil
 }
 
 func (fctx *FlowContext) deleteSubnet(ctx context.Context) error {
-	log := shared.LogFromContext(ctx)
-	current, err := fctx.findExistingSubnet()
-	if err != nil {
+	subnetID := fctx.state.Get(IdentifierSubnet)
+	if subnetID == nil {
+		return nil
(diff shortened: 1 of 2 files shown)


pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -314,7 +314,7 @@ func (fctx *FlowContext) findExistingSubnet() (*subnets.Subnet, error) {
 		return nil, err
 	}
 	if networkID == nil {
-		return nil, fmt.Errorf("network not found")
+		return nil, nil
```
AndreasBurger, line 317: This feels weird. You could also do 
```golang
		if err != nil || networkID == nil {
		return nil, err
	}
```
That would be equivalent. But I am not sure if it actually an improvement. I'll leave it up to you to decide.

AndreasBurger (approved):

pkg/controller/infrastructure/infraflow/reconcile.go
kon-angelo: Some context then:
- in this function we are looking to find an existing subnet. Some lines above we are trying to figure out the network (as we need that to find our subnet). If can't find a network, well we won't find a subnet.

That being said, I don't think that we should treat the lack of network as an error, or simply as a testament to the lack of a subnet existing (which is what this PR tries to do)

AndreasBurger: Yes. The weirdness (to me) is the explicit handling of this case. Ideally, something like this might possibly be more concisely (yet still explicitly) handled with some ignoreNotFound()-style handling, though that would require some more restructuring (if at all possible).

The change suggested above will not treat the absent networkID as an error, so I'm not sure if that is what you're getting at. But I guess it shows that it is too implicit.

kon-angelo: > Yes. The weirdness (to me) is the explicit handling of this case.

The error that prevents deletion in https://github.com/gardener/gardener-extension-provider-openstack/issues/890 is coming from the line 317 changed above (it's an error we return and not an API call error). 

In this case I can't make a an API call to get the subnet for I know not the network ID. We essentially recognize that we can't make an API call that would make sense (not without a networkID) and exit early. 

The only change here is whether this is treated as an error.

hebelsan: Looking at the places where the func `findExistingSubnet` is used:
- two times in `delete.go`, where with this PR we make sure to only delete subnet or ensure it's ID if a network is found
- one time in `reconcile.go` (ensureSubnet) where also the task dependency ensures that a network must exist

So for me, returning the error as it was before does make sense since we should never reach this case

kon-angelo: > So for me, returning the error as it was before does make sense since we should never reach this case

Well, the linked issue we try to fix does contain at least one case where we reached that error 😅. In a partial delete for example, we may have deleted the subnet and network and retry the deletion. At which point we will reach that error.

In the reconcile, you won't reach this part, because the `ensureNetwork` step is guaranteed to have finished successfully before reconciling the subnet

kon-angelo: We could probably skip the `findExistingSubnet` the second time in the deletion, but still if you can't find a network, you won't find a subnet - so I am not sure if this should really be an error.

hebelsan: ah my bad, right, not finding the subnet ID is not treated as an error....
forget my comment above

hebelsan (approved):

pkg/controller/infrastructure/infraflow/delete.go
```diff
@@ -96,48 +103,47 @@ func (fctx *FlowContext) buildDeleteGraph() *flow.Graph {
...
+		return nil
 	}
-	if current != nil {
-		log.Info("deleting...", "router", current.ID)
-		if err := fctx.networking.DeleteRouter(current.ID); err != nil {
-			return err
-		}
+
+	shared.LogFromContext(ctx).Info("deleting...", "router", *routerID)
+	if err := fctx.networking.DeleteRouter(*routerID); err != nil {
```
MichaelEischer, line 112: This refactoring leads to a new corner case if the router is deleted but the routerID is still contained in the state.

In the old version, `findExistingRouter` would return `nil` and thus the router deletion is just skipped. With the new variant, if the routerID is still part of the state (e.g. router deletion completed, but the call ran into a timeout and thus didn't clear the state) then then `DeleteRouter` call would run into an error. In my opinion, the error should be filtered with `client.IgnoreNotFoundError(err)`.

kon-angelo: Will look into it. Thank you @MichaelEischer

PR #284, opened 2021-05-19
[ci:component:github.com/gardener/machine-controller-manager-provider-openstack:v0.2.0->v0.3.0]

gardener-robot-ci-1:
**Release Notes*:
``` other dependency github.com/gardener/machine-controller-manager-provider-openstack #24 @AxiomSamarth
Revendors MCM dependent libraries for `v0.39.0` version.
```

PR #917, opened 2024-11-29
fix(openstack/networking): filter router interfaces by all possible owners (#284)

crigertg:

/area networking
/kind bug
/platform openstack

Depending on how the infrastructure in openstack was created the router port (interface) may have different owner references. A list of possible values for router interfaces can be found in the [openstack SDK](https://github.com/openstack/openstacksdk/tree/master) source code: https://github.com/openstack/openstacksdk/blob/4.0.1/openstack/cloud/_network.py#L2133

Currently only ports with owner `network:router_interface` are returned when calling `GetRouterInterfacePort`. If a subnet does not have a port with the `network:router_interface` owner for the associated router, flow attempts to create a new port. However, this creation fails because a router port for the subnet already exists.

**Release note**:

```bugfix operator
management of the router interface missed some of openstack's owner labels assigned to the routers network interface causing the infrastructure conciliation to fail due to dublicated router network interfaces
```

### DIFF
--- a/pkg/openstack/client/networking.go
+++ b/pkg/openstack/client/networking.go
@@ -7,6 +7,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/external"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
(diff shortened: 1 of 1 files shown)


pkg/openstack/client/networking.go
```diff
@@ -286,8 +286,18 @@ func (c *NetworkingClient) GetRouterInterfacePort(routerID, subnetID string) (*p
...
-	if len(list) == 0 {
+
+	validDeviceOwners := []string{
+		"network:router_interface",
+		"network:router_interface_distributed",
+		"network:ha_router_replicated_interface",
+	}
+	filtered := slices.DeleteFunc(list, func(p ports.Port) bool {
+		return !slices.Contains(validDeviceOwners, p.DeviceOwner)
+	})
```
kon-angelo, line 297: lgtm

Out of curiosity, I have to ask why did you go for the delete compared to something like
```
	for _, v := range list {
		if slices.Contains(validDeviceOwners, v.DeviceOwner) {
			return &v, nil
		}
	}
	return nil, nil
```

crigertg: tbh in the moment when fixing this I found it to be more elegant. I'm not too familiar with go yet and I was searching the `slices` package for something like `filter` in python or `select` in ruby.

kon-angelo (approved):
/lgtm

PR #918, opened 2024-11-29
configure HA webhook according to replicas in the helm chart

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

Allows you to run the extension with a single replica. For 2+ replicas it should be no-op.

Fixes #

**Release note**:

```other operator
Adjust replica count from high-availability webhook mutation according to the helm chart.
```


hebelsan (dismissed):
/lgtm

MartinWeindel (changes requested):
Looking into the code of the resource manager webhook, it will not work as expected.
See inline comment.

charts/gardener-extension-provider-openstack/templates/deployment.yaml
```diff
@@ -3,10 +3,11 @@ kind: Deployment
...
-{{-  if .Values.ignoreResources }}
   annotations:
+    {{-  if .Values.ignoreResources }}
     resources.gardener.cloud/ignore: "true"
-{{- end }}
+    {{- end }}
+    high-availability-config.resources.gardener.cloud/replicas: {{ .Values.replicaCount }}
   labels:
 {{ include "labels" . | indent 4 }}
     high-availability-config.resources.gardener.cloud/type: server
```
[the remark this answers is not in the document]
MartinWeindel: As discussed off-line, this alternative relies on some detailed behaviour of the HA webhook, which makes it hard to understand.
Maybe an boolean flag to set the annotation `high-availability-config.resources.gardener.cloud/replicas: "1"` is a better solution for the moment.

AndreasBurger (dismissed):

charts/gardener-extension-provider-openstack/templates/deployment.yaml
```diff
@@ -3,10 +3,13 @@ kind: Deployment
...
   namespace: {{ .Release.Namespace }}
-{{-  if .Values.ignoreResources }}
   annotations:
+    {{-  if .Values.ignoreResources }}
     resources.gardener.cloud/ignore: "true"
-{{- end }}
+    {{- end }}
+    {{-  if .Values.ignoreHAReplicas }}
+    high-availability-config.resources.gardener.cloud/replicas: "1"
+    {{- end }}
```
matthias-horne, line 12: if both are false it will just leave `annotation:`, which is in invalid yaml. I think we need one more if statement to catch that case.

kon-angelo: I am not sure if the empty annotations field would be an actual problem but nonetheless I applied your suggestion to the latest commit. PTAL :)

matthias-horne (approved):

PR #936, opened 2024-12-17
filter terraform output routerIP for v4 addresses only

### LINKED ISSUE #897: `status.networking.egressCIDRs[0]` is invalid: `must be valid canonical CIDR`

/area networking
/kind bug
/platform openstack

**What happened**:
After updating to Gardener 1.107.0 we get the following message:
```
task "Waiting until shoot infrastructure has been reconciled" failed: Shoot.core.gardener.cloud "shoot" is invalid: status.networking.egressCIDRs[0]: Invalid value: "2a13:1a81:8000::5d/32": must be valid canonical CIDR
```

You should not choose the first IP that comes along. You should check which IP is the correct one. For example, we do not support IPv6 in our Gardener. However, the Router provides an IPv6. Accordingly, you should check whether the IP works or is correct.

Upgrade to 1.107.0 and check whether a Router specifies an IPv6 in the first position. 

**Environment**:

- Gardener version (if relevant): 1.107.0
- Extension version: 1.42.1
- Kubernetes version (use `kubectl version`): 1.29.10 / 1.30.6
- Cloud provider or hardware configuration: Own Datacenter
- Others:

lotharbach:

/area networking
/kind bug
/platform openstack

As described in the issue, openstack routers expose a list of external IPs, and those could be v6 even if the rest of the networking stack is v4 only. This will filter terraform output `routerIP` for v4 addresses only and return only one of them just in case there could still be multiple, since the rest of the code currently expects a single v4 IP

Fixes #897

I purposefully avoided changing the API and properly handling and forwarding lists of IPs for now. I expect there is a proper v6/dualstack PR coming anyway that will need to handle this cleanly. This minimal change unblocks us for further gardener upgrades, so I wanted to share it. I can see why you would not want to merge it in favor of a proper solution. I'm happy to give this PR a little bit more love with feedback, but I don't want to waste anybodies time if this is not the right direction long-term, which I could totally understand.

**Release note**:

```other operator

```

### DIFF
--- a/pkg/internal/infrastructure/templates/main.tpl.tf
+++ b/pkg/internal/infrastructure/templates/main.tpl.tf
@@ -227,6 +227,12 @@ openstack_networking_network_v2.cluster.name
 data.openstack_networking_network_v2.cluster.name
 {{ end -}}
 {{- end -}}
+
+// Openstack API could return an IPv6 external IP even for v4-only clusters see
+// https://github.com/gardener/gardener-extension-provider-openstack/issues/897
+// We filter for IPs that match the v4 pattern (4 numbers seperated by dots)
+// and only return one just in case there are mulitple as the rest of the
+// codebase is currently expecting a single v4 IP
 {{- define "router-ip" -}}
 {{ if .create.router -}}
 one([for external_fixed_ip in openstack_networking_router_v2.router.external_fixed_ip : external_fixed_ip.ip_address if can(regex("^(?:[0-9]{1,3}\\.){3}[0-9]{1,3}$", external_fixed_ip.ip_address))])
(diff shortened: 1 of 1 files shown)


gardener-robot-ci-1:
Thank you @lotharbach for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

hebelsan:
Could you please add a comment in the template?

kon-angelo:
Hi @lotharbach. I created https://github.com/gardener/gardener-extension-provider-openstack/pull/958 to handle the multiple-ips/ipv6 case in a more similar way that we do in other extensions. Please have a look.

In the meantime I will put this PR on hold. I am sorry for the trouble/delay. I would also appreciate if you could shortly test it, since I our testing infrastructure does not have ipv6 capabilities and it would be nice to know if it actually solves your issue. 

/hold

PR #951, opened 2025-01-13
Update gophercloud client to v2 

hebelsan:

/area quality
/kind enhancement
/platform openstack

This PR updates the gophercloud client to v2.

Fixes #

**Release note**:

```feature developer
Updates gophercloud client to v2
```


hebelsan:
the infra test work with basic credentials, but something is still missing to make the client working with app credentials (what the tm is using).

pkg/openstack/client/client.go
```diff
@@ -47,36 +49,28 @@ func NewOpenstackClientFromCredentials(credentials *os.Credentials) (Factory, er
...
-
-	provider, err := openstack.NewClient(authOpts.IdentityEndpoint)
-	if err != nil {
-		return nil, err
-	}
-
-	provider.HTTPClient = *opts.HTTPClient
-
-	err = openstack.Authenticate(provider, *authOpts)
+	err = openstack.Authenticate(ctx, provider, authOpts)
```
Bobi-Wan, line 73: FYI openstack.Authenticate is called in config.NewProviderClient as well.
I managed to get a working client without the extra Authenticate for my usecase.

pkg/controller/bastion/actuator_reconcile.go
```diff
@@ -322,31 +323,57 @@ func addressToIngress(dnsName *string, ipAddress *string) *corev1.LoadBalancerIn
...
+
+	if fip.ID != "" {
+		return nil
+	}
+
+	if fip.Status != "ACTIVE" || instance.Status != "ACTIVE" {
+		return fmt.Errorf("instance or floating ip address not ready yet")
+	}
+
+	return networkClient.UpdateFIPWithPort(ctx, floatingIP.ID, instancePort.ID)
```
kon-angelo, line 373: I am not sure I understand this part. You already associate the fip if it's not already. Why do you make a second update call ?

pkg/controller/bastion/actuator_reconcile.go
```diff
@@ -322,31 +323,57 @@ func addressToIngress(dnsName *string, ipAddress *string) *corev1.LoadBalancerIn
...
+			return err
+		}
+		fip = *floatingIP
+	}
+
+	if fip.ID != "" {
+		return nil
+	}
+
+	if fip.Status != "ACTIVE" || instance.Status != "ACTIVE" {
```
kon-angelo, line 369: Maybe something to doublecheck, but I believe the FIP should already be active upon successful creation, and well, likewise for the instance. Is assigning the IP to the instance actually changing the instance status or the FIP status ? I would also argue that if not then there could be a wait introduced but not in this place.

pkg/controller/bastion/actuator_reconcile.go
hebelsan: True it's associated twice which doesn't make sense.

pkg/controller/bastion/actuator_reconcile.go
hebelsan: It's true, this logic doesn't make a lot of sense.
Especially because even if the status does change, we would need to retrieve the object again.
I'm in favour of introducing the status check in the reconcile func with a retry like you suggested.

kon-angelo (approved):
/lgtm

PR #956, opened 2025-01-15
fix config validator for region in network client

### LINKED ISSUE #955: config validation for network fails in other regions that are not default

/area robustness
/area networking
/kind bug
/platform openstack

**What happened**: Tried creating shoot in other regions than the default with OpenStack configuration of our infrastructure, but it refused to discover network from the selected region.

**What you expected to happen**: Healthy creation of infrastructure.

**Anything else we need to know?**: I have also created and tested a fix for this issue. Currently while creating networking client for terraform validations, region is not being used as parameter for the function.

**Environment**:

- Extension version: v1.43.2
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration: OpenStack
- Others:

snwzd:
Added region argument while creating network client for terraform validations

/area robustness
/kind bug
/platform openstack

Fixes #955 

While creating networking client for terraform validations, region is not being used as parameter for the function.


gardener-robot-ci-2:
Thank you @snwzd for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

snwzd:
Hi, This pull request is regarding the fix for the issue issue that I am facing, please review it and tell me if there are any other changes that need to be made.

pkg/controller/infrastructure/configvalidator.go
```diff
@@ -61,7 +62,10 @@ func (c *configValidator) Validate(ctx context.Context, infra *extensionsv1alpha
 		allErrs = append(allErrs, field.InternalError(nil, fmt.Errorf("could not create Openstack client factory: %+v", err)))
 		return allErrs
 	}
-	networkingClient, err := clientFactory.Networking()
+	networkingClient, err := clientFactory.Networking(func(opts gophercloud.EndpointOpts) gophercloud.EndpointOpts {
+		opts.Region = infra.Spec.Region
+		return opts
+	})
```
kon-angelo, line 68: We already should have a function to do just that, so can you give it a try with:

```
	networkingClient, err := clientFactory.Networking(openstackclient.WithRegion(infra.Spec.Region))
```

kon-angelo (commented):
@snwzd thank you for the contribution. Only a minor nit below

kon-angelo (dismissed):
/reviewed ok-to-test

pkg/controller/infrastructure/configvalidator.go
kon-angelo: @snwzd you need to run `make generate` (according to the test results)

snwzd: I ran make generate, but it didn’t seem to make much changes. I think it removed the unused gophercloud import, which was the only noticeable change. (probably my editor missed it)

kon-angelo:
@snwzd I am not sure what was the issue with the previous run of the test, but now it seems to fail at the unit tests level, because the call to the client interface is missing the new argument (region). 

Can you please update the tests to use the region too ? For that matter, are you able to view the test logs from the verify pipeline ?

AndreasBurger (approved):

kon-angelo:
@snwzd thank you for the contribution

PR #970, opened 2025-01-31
fix(infraflow): report last error for wait functions

hown3d:

/area quality
/kind enhancement
/platform openstack

Adds the last error seen for Wait functions to the returned error if a context is done. 

If a user adds servers into the network created by the extension and attaches a floating IP to it, openstack will report with an 409 error which is explicitly filtered out in the functions. If a timeout occurs and the context is canceled, infraflow will only report context canceled in the infrastructure `status.lastError` field. Therefore is hard to understand why the task failed in the first place.

Example of the new error format:
```yaml
status:
  lastError:
    description: "Error deleting Infrastructure: 1 error occurred:\n\t* failed to
      \"delete router interface\": context deadline exceeded, last error: Expected
      HTTP response code [200] when accessing [PUT https://XXXX:443/v2.0/routers/1b3a04d1-49b1-494f-b7e1-edf23f86ad8b/remove_router_interface],
      but got 409 instead\n{\"NeutronError\": {\"type\": \"RouterInterfaceInUseByFloatingIP\",
      \"message\": \"Router interface for subnet XXXX
      on router XXXX cannot be deleted, as it is required
      by one or more floating IPs.\", \"detail\": \"\"}}\n\n"
```

Replace `time.Sleep` functions with a `time.Ticker` and use select.

**Release note**:

```bugfix operator
infraflow: report last error on task timeouts
```

### DIFF
--- a/pkg/controller/infrastructure/infraflow/access/networking_access.go
+++ b/pkg/controller/infrastructure/infraflow/access/networking_access.go
@@ -204,37 +204,38 @@ func (a *networkingAccess) AddRouterInterfaceAndWait(ctx context.Context, router
 	if err != nil {
 		return err
 	}
+	ticker := time.Tick(3 * time.Second)
 	for {
-		if ctx.Err() != nil {
-			return ctx.Err()
-		}
-		time.Sleep(1 * time.Second)
-		port, err := a.networking.GetPort(info.PortID)
-		if err != nil {
-			return err
-		}
-
-		if port == nil {
-			return fmt.Errorf("port was not found ")
-
-		}
-
-		switch port.Status {
-		case "BUILD", "PENDING_CREATE", "PENDING_UPDATE", "DOWN":
-			a.log.Info("port is not in expected state", "Port", info.PortID, "Status", port.Status)
-			if a.log.V(1).Enabled() {
-				marshalled, err := json.Marshal(info)
-				if err != nil {
-					return err
-				}
-				a.log.V(1).Info("port info", "Port", string(marshalled))
+		select {
+		case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


gardener-robot-ci-2:
Thank you @hown3d for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

kon-angelo (approved):
/lgtm

AndreasBurger:
The current verify-error was due to the commit on master this PR was based on, should be good after a rebase.

While you're at it/in case you're up for it, the release note currently is not right (`\` added for escaping):

```
\```other operator
bugfix operator
infraflow: report last error on task timeouts
\```
```
the inner `bugfix operator` should be where `other operator` is currently.

pkg/controller/infrastructure/infraflow/access/networking_access.go
```diff
@@ -253,23 +254,26 @@ func (a *networkingAccess) GetRouterInterfacePortID(routerID, subnetID string) (
...
+					return nil
+				}
+				if _, ok := err.(gophercloud.ErrDefault409); !ok {
+					return err
+				}
 			}
-			if _, ok := err.(gophercloud.ErrDefault409); !ok {
-				return err
+			if err == nil {
+				return nil
```
kon-angelo, line 274: Does this look better:

```
				if _, ok := err.(gophercloud.ErrDefault409); !ok {
					return err
				}
                continue
            }
            return nil
```

It's not entirely on this PR for the style change but it looks more readable in my opinion

kon-angelo (approved):
/lgtm

PR #975, opened 2025-02-05
[Release v1.45] add max-entries to attacher

AndreasBurger:

/area control-plane
/kind bug
/platform openstack

Adds a limit to the listVolume call made by the attacher. The csi-driver call to cinder may return only a limited amount of volumes per page and without the max-entries these additional volumes will not be processed.

Fixes #

**Release note**:

```other operator
Add max-entries to the csi-attacher arguments. 
```


kon-angelo (approved):
/lgtm

PR #994, opened 2025-03-03
Fix Deployment in Garden Runtime Cluster

timuthy:

/area delivery
/kind bug
/platform openstack

This PR mainly fixes an issue that occurred when deploying the extension as an Gardener Operator extension to the garden runtime cluster. 
Because of the seed default values, it unconditionally deployed the `gardener.cloud-fast` storage class which is not only unexpected if the runtime garden is no OpenStack cluster, but caused race conditions with other provider extensions in the very same cluster (see commit descriptions).

/cc @MartinWeindel

**Release note**:

```bugfix operator
Fixed an issue that caused deployment issues with the `gardener.cloud-fast` storage class when the extension was deployed by `gardener-operator` in the garden runtime cluster. The deployment of this `StorageClass` object is now only done for OpenStack seeds.
```


kon-angelo (approved):
/lgtm

PR #995, opened 2025-03-04
fix(worker): select computeClient based on region

### LINKED ISSUE #801: Broken network/compute/dns clients if multiple openstack regions exist

/area networking, os, storage
/kind bug
/platform openstack

**What happened**:

When multiple regions exist in the OpenStack service catalog, then the gardener-extension-provider-openstack may fail to create a working client to access the compute / network / dns service of openstack. The clients use the first public endpoint in the OpenStack service catalog, which might be the correct one or not. This results in a complete failure to create machines, networks, DNS etc.

As far as I can tell, only the terraform based flow is affected, not the new infraflow variant. See below for a Proof of Concept fix.

Communication with the OpenStack API should use the correct endpoints.

I unfortunately don't have an easy way to reproduce. Essentially you need an openstack installation, then add a second region to the service catalog (some placeholder entries for compute, network and dns should work, as long as their name is stored before the actual region). The buggy client instances will use the first region returned by the service catalog, independent of where the actual resources are located.

There is a proof of concept fix at [here](https://github.com/stackitcloud/gardener-extension-provider-openstack/commit/82c4e736a68ae3908d9abff20cb3236d1832ca9c)

**Environment**:

- Extension version: 1.38.2 / 1.39.2 (issue also affects the latest master branch)
- Kubernetes version (use `kubectl version`): 1.28
- Cloud provider or hardware configuration: OpenStack with multiple regions in the service catalog

mreichardt95:

/area os
/kind bug
/platform openstack

Worker controller should select compute endpoints based on Region instead of picking the first `public` endpoint.

Fixes #801 

**Release note**:

```bugfix operator
Correctly select endpoints in multi-region OpenStack environments
```

### DIFF
--- a/pkg/controller/worker/machine_dependencies.go
+++ b/pkg/controller/worker/machine_dependencies.go
@@ -30,7 +30,7 @@ func (w *workerDelegate) CleanupMachineDependencies(_ context.Context) error {
 
 // PreReconcileHook implements genericactuator.WorkerDelegate.
 func (w *workerDelegate) PreReconcileHook(ctx context.Context) error {
-	computeClient, err := w.openstackClient.Compute()
+	computeClient, err := w.openstackClient.Compute(osclient.WithRegion(w.worker.Spec.Region))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


gardener-robot-ci-2:
Thank you @mreichardt95 for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

kon-angelo (approved):
/lgtm

PR #1047, opened 2025-05-06
Set user agent when creating OpenStack client

AndreasBurger:

/area audit-logging
/kind enhancement
/platform openstack

Adds information to the UserAgent string of the created client to make it easier to identify calls made by clients created by the extension provider.

Fixes #

**Release note**:

```other operator
OpenStack-clients created by the OpenStack extension provider will now identify themselves by adding to the `user-agent` header of their calls.
```

### DIFF
--- a/pkg/openstack/client/client.go
+++ b/pkg/openstack/client/client.go
@@ -70,6 +70,8 @@ func NewOpenstackClientFromCredentials(ctx context.Context, credentials *os.Cred
 		return nil, err
 	}
 
+	provider.UserAgent.Prepend("Gardener Extension for OpenStack provider")
+
 	return &OpenstackClientFactory{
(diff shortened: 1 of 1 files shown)


pkg/openstack/client/client.go
```diff
@@ -70,6 +70,8 @@ func NewOpenstackClientFromCredentials(ctx context.Context, credentials *os.Cred
 		return nil, err
 	}
 
+	provider.UserAgent.Prepend("Gardener Extension for OpenStack provider")
```
kon-angelo, line 73: What about `gardener-extension-provider-openstack`

Otherwise lgtm

kon-angelo (approved):
/lgtm

PR #1060, opened 2025-05-22
Setting `internal server error` to the right KnownCodes group

etiennnr:

/area ops-productivity
/kind bug post-mortem
/platform openstack

Changing the `internal server error` error message to be retriable. Basically we had an issue where there was an update to the Neutron API, which caused such HTTP 500 `internal server error`. Thus, this was transient (OpenStack API was back working just a few minutes later). However, the shoot was stuck in a `user error` which required manual reconcile triggering.

The only potential use case that could see where this might cause a bigger problem is if there's a wide / prolonged outage of OpenStack API. However, I believe [gardenlet already has protections in place](https://github.com/gardener/gardener/blob/master/pkg/gardenlet/controller/shoot/shoot/reconciler.go) to stop retrying if too many attempts fail (feel free to correct me if I'm wrong). Thus, I don't see why this kind of error should be not retriable.

Fixes #

**Release note**:

```bugfix bugfix
Shoot reconcile will retry if we get the `internal server error` message from Openstack
```


gardener-robot-ci-1:
Thank you @etiennnr for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

kon-angelo (approved):
/lgtm

hebelsan (approved):
/lgtm

PR #1074, opened 2025-06-09
Tag case where floating IPs prevent router interface deletion as user error

Kostov6:

/area ops-productivity
/kind enhancement
/platform openstack

It seems the extension does not create or delete any floating IP addresses in the shoot subnet so any floating IPs that are there and prevent router interface deletion are user created floating IPs

**Release note**:

```noteworthy user
Floating IPs in the shoot subnet preventing router deletion is tagged as `ERR_INFRA_DEPENDENCIES` user error
```


shafeeqes:
Can we merge this as well ? @hebelsan 
Can you please rebase @Kostov6 ?

kon-angelo:
A FIP can be created in response to LBs, bastions and the likes. Sure when the shoot is getting deleted then these resources should be cleaned but there are conditions that prevents this from happening. 

Do we want to assign this as a user-error as the more prevalent case ?

Kostov6:
As bastions created by gardener are on the same shoot subnet leaked floating ip can also be a gardener error

PR #1076, opened 2025-06-12
Remove NetworkInUse from KnownErrors

hebelsan:

/area usability
/kind bug
/platform openstack

This PR reverts https://github.com/gardener/gardener-extension-provider-openstack/pull/1049 because a leaked port and a non-shoot server in shoot network scenario have the same error message.
Therefore we should not use `NetworkInUse` as KnwonError regex string.

Fixes #

**Release note**:

```other operator
NONE
```


kon-angelo (approved):
/lgtm

PR #1085, opened 2025-06-24
Fix failure because of race conditions for `update-resolv-conf.service`

### LINKED ISSUE #782: Race condition while generating `resolv-for-kubelet.conf`

/area robustness
/kind bug
/platform openstack

**What happened**:
`/etc/resolv-for-kubelet.conf` was generated without any `nameserver` in it.
Observing the (redacted) logs:
```
2024-05-27T08:45:12.168190+0000 node-name systemd[1]: Starting update-resolv-conf.service - update /etc/resolv-for-kubelet.conf on start and after each change of /run/systemd/resolve/resolv.conf...
2024-05-27T08:45:12.192931+0000 node-name update-resolv-conf.sh[2138034]: updated /etc/resolv-for-kubelet.conf
2024-05-27T08:45:12.193163+0000 node-name systemd[1]: update-resolv-conf.service: Deactivated successfully.
2024-05-27T08:45:12.193349+0000 node-name systemd[1]: Finished update-resolv-conf.service - update /etc/resolv-for-kubelet.conf on start and after each change of /run/systemd/resolve/resolv.conf.
```
and when checked:
```bash
root@node-name:/# ls -la --full /run/systemd/resolve/resolv.conf
-rw-r--r-- 1 systemd-resolve systemd-resolve 833 2024-05-27 08:45:12.188413386 +0000 /run/systemd/resolve/resolv.conf

root@node-name:/# ls -la --full /etc/resolv*
-rw-r--r-- 1 root root 906 2024-05-27 08:45:12.168411571 +0000 /etc/resolv-for-kubelet.conf
lrwxrwxrwx 1 root root  32 2023-11-03 00:00:00.000000000 +0000 /etc/resolv.conf -> /run/systemd/resolve/resolv.conf
```

`/etc/resolv-for-kubelet.conf` to always be generated with the latest data.

I see that we are handling this in https://github.com/gardener/gardener-extension-provider-openstack/blob/master/pkg/webhook/controlplane/ensurer.go#L390
We are generating a shell script that reacts on being called by a one shot service on path changes.
This works nicely on a file change but fails if called multiple times while systemd is processing the spawn of that process.
Issue stems from the way systemd will handle multiple calls from the same time if received for the state we are already in.

**Environment**:

- Extension version: v1.39.2
- Kubernetes version (use `kubectl version`): v1.28.6
- Cloud provider or hardware configuration: Openstack
- Others:

RadaBDimitrova:

/area ops-productivity
/kind bug
/platform openstack

The starting of `update-resolv-conf.path` can happen after the update of `/run/systemd/resolve/resolv.conf` with the valid configuration. The following has been observed where the file has been updated 4 sec before the start of `update-resolv-conf.path`:
```
cat /etc/resolv-for-kubelet.conf
# This is /run/systemd/resolve/resolv.conf managed by man:systemd-resolved(8).
# Do not edit.
#
# This file might be symlinked as /etc/resolv.conf. If you're looking at
# /etc/resolv.conf and seeing this text, you have followed the symlink.
#
# This is a dynamic resolv.conf file for connecting local clients directly to
# all known uplink DNS servers. This file lists all configured search domains.
#
# Third party programs should typically not access this file directly, but only
# through the symlink at /etc/resolv.conf. To manage man:resolv.conf(5) in a
# different way, replace this symlink by a static file or a different symlink.
#
# See man:systemd-resolved.service(8) for details about the supported modes of
# operation for /etc/resolv.conf.

# No DNS servers known.
search .

# updated by update-resolv-conf.service (installed by gardener-extension-provider-openstack)
options rotate timeout:1
```
instead of:
```
cat /etc/resolv.conf
# This is /run/systemd/resolve/resolv.conf managed by man:systemd-resolved(8).
# Do not edit.
#
# This file might be symlinked as /etc/resolv.conf. If you're looking at
# /etc/resolv.conf and seeing this text, you have followed the symlink.
#
# This is a dynamic resolv.conf file for connecting local clients directly to
# all known uplink DNS servers. This file lists all configured search domains.
#
# Third party programs should typically not access this file directly, but only
# through the symlink at /etc/resolv.conf. To manage man:resolv.conf(5) in a
# different way, replace this symlink by a static file or a different symlink.
#
# See man:systemd-resolved.service(8) for details about the supported modes of
# operation for /etc/resolv.conf.

nameserver 10.191.8.2
nameserver 10.191.8.3
search <name>

# updated by update-resolv-conf.service (installed by gardener-extension-provider-openstack)
options rotate timeout:1
```
This can lead to calico and other components' downtime because of faulty DNS resolution.

Fixes #782 

/cc @Kostov6 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -239,6 +239,8 @@ StartLimitIntervalSec=0
 [Service]
 Type=oneshot
 ExecStart=/opt/bin/update-resolv-conf.sh
+RestartForceExitStatus=10
+RestartSec=15
 `
 	)
 
@@ -283,6 +285,10 @@ is_systemd_resolved_system()
       return 1
     fi
 }
+
+if grep -q "# No DNS servers known." /run/systemd/resolve/resolv.conf; then
+  exit 10
+fi
 
 rm -f "$tmp"
 if is_systemd_resolved_system; then
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -239,7 +239,7 @@ StartLimitIntervalSec=0
 [Service]
 Type=oneshot
 ExecStart=/opt/bin/update-resolv-conf.sh
-RestartForceExitStatus=10
+RestartForceExitStatus=78
 RestartSec=15
 `
 	)
@@ -287,7 +287,7 @@ is_systemd_resolved_system()
 }
 
 if ! grep -Eq "^nameserver\s+" /run/systemd/resolve/resolv.conf; then
-  exit 10
+  exit 78
 fi
 
 rm -f "$tmp"
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -286,7 +286,7 @@ is_systemd_resolved_system()
     fi
 }
 
-if grep -q "# No DNS servers known." /run/systemd/resolve/resolv.conf; then
+if ! grep -Eq "^nameserver\s+" /run/systemd/resolve/resolv.conf; then
   exit 10
 fi
 
--- a/pkg/webhook/controlplane/ensurer_test.go
+++ b/pkg/webhook/controlplane/ensurer_test.go
@@ -56,7 +56,7 @@ is_systemd_resolved_system()
 }
 
 if ! grep -Eq "^nameserver\s+" /run/systemd/resolve/resolv.conf; then
-  exit 10
+  exit 78
 fi
 
 rm -f "$tmp"
@@ -354,7 +354,7 @@ StartLimitIntervalSec=0
 [Service]
 Type=oneshot
 ExecStart=/opt/bin/update-resolv-conf.sh
-RestartForceExitStatus=10
+RestartForceExitStatus=78
 RestartSec=15
 `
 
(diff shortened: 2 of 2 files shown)


Kostov6:
@MartinWeindel do you remember why a `oneshot` unit type was chosen instead of `simple` in https://github.com/gardener/gardener-extension-provider-openstack/pull/342?

RadaBDimitrova:
/unhold
it's been validated with a local extension setup

MartinWeindel:
> @MartinWeindel do you remember why a `oneshot` unit type was chosen instead of `simple` in #342?

The assumption was that it is sufficient to configure it once at boot time. Probably I just adapted from an example.

hebelsan (approved):
/lgtm

pkg/webhook/controlplane/ensurer.go
```diff
@@ -239,6 +239,8 @@ StartLimitIntervalSec=0
 [Service]
 Type=oneshot
 ExecStart=/opt/bin/update-resolv-conf.sh
+RestartForceExitStatus=10
```
dnaeon, line 242: A better approach would be to pick a standard exit code such as `EX_CONFIG (78)` here instead of `10`.

See [sysexits(3)](https://man7.org/linux/man-pages/man3/sysexits.h.3head.html) for more details.

kon-angelo (approved):
/lgtm

PR #1155, opened 2025-09-17
add shoot input validation

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

Fixes #

**Release note**:

```other operator
Add shoot input validation
```

### DIFF
--- a/pkg/apis/openstack/validation/filter.go
+++ b/pkg/apis/openstack/validation/filter.go
@@ -44,7 +44,7 @@ func regex(regex string) validateFunc[string] {
 			return allErrs
 		}
 		if !compiled.MatchString(name) {
-			allErrs = append(allErrs, field.Invalid(fld, name, fmt.Sprintf("does not match expected regex %s", compiled.String())))
+			allErrs = append(allErrs, field.Invalid(fld, name, fmt.Sprintf("does not match expected regex %q", compiled.String())))
 		}
 		return allErrs
 	}
@@ -52,7 +52,7 @@ func regex(regex string) validateFunc[string] {
 
 func notEmpty(name string, fld *field.Path) field.ErrorList {
 	if utf8.RuneCountInString(name) == 0 {
-		return field.ErrorList{field.Required(fld, name)}
+		return field.ErrorList{field.Required(fld, "cannot be empty")}
 	}
 	return nil
 }
--- a/pkg/apis/openstack/validation/infrastructure.go
+++ b/pkg/apis/openstack/validation/infrastructure.go
@@ -64,7 +64,7 @@ func ValidateInfrastructureConfig(infra *api.InfrastructureConfig, nodesCIDR *st
 		allErrs = append(allErrs, validateResourceName(*infra.FloatingPoolSubnetName, fldPath.Child("floatingPoolSubnetName"))...)
 
 		if infra.Networks.Router != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("floatingPoolSubnetName"), *infra.FloatingPoolSubnetName, "router id must be empty when a floating subnet name is provided"))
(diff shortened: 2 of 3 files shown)


pkg/apis/openstack/validation/filter.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"regexp"
+	"unicode/utf8"
+
+	guuid "github.com/google/uuid"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+var (
+	// genericNameRegex is used to validate openstack resource names. They generally do not impose any restrictions on the characters used, but we constrain curly braces or newlines.
```
hebelsan, line 17: maybe add brackets to the comment since we also disallow them

pkg/apis/openstack/validation/filter.go
```diff
@@ -0,0 +1,65 @@
...
+// regex returns a filterFunc that validates a string against a regular expression.
+func regex(regex string) validateFunc[string] {
+	compiled := regexp.MustCompile(regex)
+	return func(name string, fld *field.Path) field.ErrorList {
+		var allErrs field.ErrorList
+		if name == "" {
+			return allErrs // Allow empty strings to pass through
+		}
+		if !compiled.MatchString(name) {
+			allErrs = append(allErrs, field.Invalid(fld, name, fmt.Sprintf("does not match expected regex %s", compiled.String())))
```
hebelsan, line 44: We compare runes with `utf8.RuneCountInString`, but report bytes via `len(name)`.

pkg/apis/openstack/validation/filter.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		return allErrs
+	}
+}
+
+func maxLength(max int) validateFunc[string] {
+	return func(name string, fld *field.Path) field.ErrorList {
+		var allErrs field.ErrorList
+		if utf8.RuneCountInString(name) > max {
```
hebelsan, line 53: also here I'd do something like:
```
l := utf8.RuneCountInString(name)
if l > max {
  allErrs = append(allErrs, field.Invalid(fld, name, fmt.Sprintf("must not be more than %d characters, got %d", max, l)))
}
```

pkg/apis/openstack/validation/cloudprofile.go
```diff
@@ -269,6 +269,20 @@ func validateLoadBalancerClass(lbClass api.LoadBalancerClass, fldPath *field.Pat
...
 		allErrs = append(allErrs, field.Invalid(fldPath, *lbClass.Purpose, fmt.Sprintf("invalid LoadBalancerClass purpose. Valid values are %q or %q", api.DefaultLoadBalancerClass, api.PrivateLoadBalancerClass)))
 	}
+	if lbClass.FloatingNetworkID != nil {
+		allErrs = append(allErrs, uuid(*lbClass.FloatingNetworkID, fldPath.Child("floatingNetworkID"))...)
+	}
+	if lbClass.FloatingSubnetID != nil {
+		allErrs = append(allErrs, uuid(*lbClass.FloatingSubnetID, fldPath.Child("floatingSubnetID"))...)
+	}
+
+	allErrs = append(allErrs, validateResourceName(lbClass.Name, fldPath.Child("name"))...)
```
hebelsan, line 279: Should we maybe also check that the `lbClass.Name` is not empty?

hebelsan (changes requested):

pkg/apis/openstack/validation/cloudprofile.go
hebelsan: doesn't the regex do this:
```
if name == "" {
  return allErrs // Allow empty strings to pass through
}
```

hebelsan (dismissed):
/lgtm

PR #1161, opened 2025-09-25
Adapt the retry logic for the resolv.conf copy script to deal with ancient systemd releases.

ScheererJ:

/area networking
/kind enhancement
/platform openstack

Adapt the retry logic for the resolv.conf copy script to deal with ancient systemd releases.

Some dated linux distributions use old systemd releases, which do not allow units to specify RestartForceExitStatus and Type=oneshot to be set at the same time. To allow the systemd unit to be executed also in those systems we use the less elegant solution using a loop.

Fixes #

The corresponding error message from a node prior to this change:

```
# journalctl -u update-resolv-conf.service
Sep 25 14:01:08 shoot--my-project--shoot-worker-r4575-z1-79dd4-j8dsl systemd[1]: update-resolv-conf.service: Service has RestartForceExitStatus= set, which isn't allowed for Type=oneshot services. Refusing.
```

**Release note**:

```other operator

```

### DIFF
--- a/pkg/webhook/controlplane/ensurer.go
+++ b/pkg/webhook/controlplane/ensurer.go
@@ -239,8 +239,6 @@ StartLimitIntervalSec=0
 [Service]
 Type=oneshot
 ExecStart=/opt/bin/update-resolv-conf.sh
-RestartForceExitStatus=78
-RestartSec=15
 `
 	)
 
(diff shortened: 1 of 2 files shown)


kon-angelo (approved):
/lgtm

PR #1166, opened 2025-09-30
Remove terraformer references

kon-angelo:

/area control-plane
/kind enhancement
/platform openstack

- removes references to the terraformer image
- only uses flow for reconciliation. Migration from terraform is still supported
- consolidates tests to a single testrun. The new test in addition to creation, also tests a second reconciliation and a recovery in case of lost state. Previously these used to be different testruns.

Fixes #

**Release note**:

```breaking operator
Deprecate usage of terraformer reconciler. Existing infrastructure objects that have yet to be migrated, will be reconciled with the flow reconciler instead.
```


hebelsan (commented):
nice work! Just two small comments.

hebelsan (commented):
Just a very small nit.

PR #1179, opened 2025-10-17
remove use-octavia field

kon-angelo:

/area control-plane
/kind technical-debt
/platform openstack

The field was intended to be used to configure the CCM, however CCM only works with octavia anymore since v1.25 which is no longer supported by gardener. The field has long been marked deprecated. Hence we can proceed with removing it altogether.

Fixes #

**Release note**:

```breaking operator
Remove use-octavia field uses from the cloudprofile and CCM charts. Operators must update their `cloudprofiles` if they still use the field.
```


hebelsan (changes requested):
Nice work!
Could you please also remove the documentation on how to enable octavia in `operations.md`?

PR #1209, opened 2025-11-20
NodeTemplate inherits attributes from pool and now supports VirtualCapacity for virtual extended resources

### LINKED ISSUE #1208: Support extended (including virtual) resources in provider config NodeTemplate without re-specifying core resources

/area auto-scaling
/kind enhancement
/platform openstack

* Permit specification of custom extended resources in the `worker.providerConfig.nodeTemplate` without needing to explicitly re-specify core resources such as `cpu`, `gpu`, `memory` for the machine type again. Basically, nodeTemplate inherits and can be overridden
*  The `worker.providerConfig.NodeTemplate` exposes an additional map that specify *virtual* extended resources of a Node that don't require rollout as opposed to *real* provisionable resources

```yaml
providerConfig:
    nodeTemplate:
        capacity:
            cpu: 123
        virtualCapacity: 
             hc.hana.com/memory: 1234567
```

**Why is this needed**:

This is needed for support of https://github.com/gardener/autoscaler/issues/132. This will permit stakeholders to have custom extended resources associated with Nodes, thus permitting Pods specs with requests for these custom extended resources and leveraging the scheduling support by the kube-scheduler

Currently, specifying an extended resources for a worker pool in the shoot specification triggers rollout of all nodes of that worker pool.  This is not semantically in-correct since extended resources in the general case are used to represent provisionable "real" resources which should necessitate a rollout.  However, "virtual" resources that are that represent some abstract/computational property of the Node should not trigger a rollout after the worker pool is updated.

elankath:
…nded resources

/area auto-scaling
/kind enhancement
/platform openstack

Fixes #1208

It appears that openstack worker pool hash is still using `v1` hash and that (unlike aws extension provider). Also, `generateWorkerPoolHash`  at https://github.com/gardener/gardener-extension-provider-openstack/blob/edd46e48dd8753df739fee334a2e8ef06bf5f460/pkg/controller/worker/machines.go#L281 apparently sets `pool.ProviderConfig = nil`, so there is never a rollout if the provider config changes. Is this understanding right ?

**Release note**:

```feature operator
Support specification of extended resources in provider config nodeTemplate.Capacity without re-specifying core resources.
Support virtual extended resources in provider config nodeTemplate.VirtualCapacity which should hot-update the node.Status.Capacity.
```

### DIFF
--- a/docs/usage/usage.md
+++ b/docs/usage/usage.md
@@ -161,9 +161,13 @@ serverGroup:
   policy: soft-anti-affinity
 # nodeTemplate: # (to be specified only if the node capacity would be different from cloudprofile info during runtime)
 #   capacity:
-#     cpu: 2
-#     gpu: 0
-#     memory: 50Gi
+#     cpu: 2 # inherited from pool's machine type if un-specified
+#     gpu: 0 # inherited from pool's machine type if un-specified
+#     memory: 50Gi # inherited from pool's machine type if un-specified
+#    ephemeral-storage: 10Gi # override to specify explicit ephemeral-storage for scale from zero
+#    resource.com/dongle: 4 # example of a custom, extended resource.
+#    virtualCapacity:
+#      subdomain.domain.com/resource-name: 1234567 # should be hot-updated in worker pool's node.Status.Capacity without rollout
 # machineLabels:
 #  - name: my-label
 #    value: foo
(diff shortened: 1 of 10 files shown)


aaronfern:
> Is this still needed with the merge of https://github.com/gardener/gardener-extension-provider-openstack/pull/1266?
cc @aaronfern

No, this PR can be closed

PR #1257, opened 2026-02-04
Add dual-stack IPv4/IPv6 support for OpenStack shoots

axel7born:
- Add SubnetPoolID field to InfrastructureConfig for IPv6 subnet allocation
- Create separate IPv6 subnets for nodes, pods, and services from subnet pool
- Add validation requiring SubnetPoolID for dual-stack configurations
- Update machine classes to support dual subnets and pod network CIDRs

/area networking
/kind enhancement
/platform openstack

This PR adds dual-stack (IPv4/IPv6) networking support for OpenStack shoots. 

The implementation creates separate IPv6 subnets for nodes, pods, and services, allocated from a configurable OpenStack subnet pool. This approach ensures proper address scope management and compatibility with external routing infrastructure.

Fixes #

**Release note**:

```other operator
Add dual-stack (IPv4/IPv6) networking support for OpenStack shoots. Configure `subnetPoolID` in InfrastructureConfig to allocate IPv6 subnets for nodes, pods, and services.
```


ScheererJ:
> By now it's not possible in the [Shoot spec](https://github.com/gardener/gardener/blob/master/docs/api-reference/core.md#core.gardener.cloud/v1beta1.Networking) to specify IPv4 and IPv6 CIDRs for node, pods and services. We would need an API change in gardener/gardener to implement this. @ScheererJ @DockToFuture @domdom82 What do you think?

It was a conscious decision to move away from providing the IP ranges for IPv6 upfront in most cases and let the infrastructure handle it. I would prefer to keep it that way.
However, it could be possible to specify the missing range(s) in the provider configuration. For other infrastructures, e.g. AWS or GCP, the IPv4 ranges are specified as before for dual-stack clusters, but the IPv6 range is provided by the infrastructure. If there is a need to manage IPv6 externally in openstack infrastructures, we can consider adding a new attribute next to `workers`.

@kon-angelo What do you think?

pkg/controller/worker/machines.go
```diff
@@ -201,7 +178,14 @@ func (w *WorkerDelegate) generateMachineConfig(ctx context.Context) error {
...
 			}
 
-			machineClassSpec["subnetID"] = subnet.ID
+			// Collect all subnet IDs
+			var subnetIDs []string
+			for _, subnet := range subnets {
+				subnetIDs = append(subnetIDs, subnet.ID)
+			}
+
+			machineClassSpec["subnetID"] = subnets[0].ID
```
hebelsan: `subnetID` is deprecated and we already have the information in `subnetIDs`, therefore I'd remove it.

charts/internal/machineclass/templates/machineclass.yaml
```diff
@@ -57,7 +54,10 @@ providerSpec:
 {{- end }}
     networkID: {{ $machineClass.networkID }}
     subnetID: {{ $machineClass.subnetID }}
```
hebelsan: `subnetID` is deprecated and we already have the information in `subnetIDs`, therefore I'd remove it.

docs/usage/usage.md
```diff
@@ -87,6 +87,45 @@ Apart from the router and the worker subnet the OpenStack extension will also cr
...
+subnetPoolID: MY-SUBNET-POOL-ID  # IPv6 subnet pool
+networks:
+  workers: 10.250.0.0/19
+```
+
+Unlike IPv4 where you explicitly specify the CIDR in `networks.workers`, IPv6 addresses are typically allocated from a larger prefix managed by your OpenStack environment.
+
+#### Option 2: Explicit IPv6 CIDR Configuration
+
+Alternatively, you can explicitly specify IPv6 CIDRs for nodes, pods, and services using the `networks.ipv6` field:
```
hebelsan: In this case it will allocate IPv6 subnets from the default subnet pool right?
I'd add this info also here in the guide.

pkg/admission/validator/shoot.go
```diff
@@ -169,19 +167,25 @@ func (s *shoot) validateShootUpdate(ctx context.Context, oldShoot, shoot *core.S
...
 
-func (s *shoot) validateShoot(ctx context.Context, context *validationContext) field.ErrorList {
+func (s *shoot) validateShoot(context *validationContext) field.ErrorList {
 	allErrs := field.ErrorList{}
 	if context.shoot.Spec.Networking != nil {
 		allErrs = append(allErrs, openstackvalidation.ValidateNetworking(context.shoot.Spec.Networking, nwPath)...)
 		allErrs = append(allErrs, openstackvalidation.ValidateInfrastructureConfig(context.infraConfig, context.shoot.Spec.Networking.Nodes, infraConfigPath)...)
+
+		// Validate that either SubnetPoolID or IPv6 config is set for dual-stack shoots
+		if !core.IsIPv4SingleStack(context.shoot.Spec.Networking.IPFamilies) {
```
hebelsan: If you rebase to master you should be able to use the api func `IsDualStack()` from g/g.

pkg/admission/validator/shoot.go
```diff
@@ -169,19 +167,25 @@ func (s *shoot) validateShootUpdate(ctx context.Context, oldShoot, shoot *core.S
...
+			if context.infraConfig.SubnetPoolID == nil && context.infraConfig.Networks.IPv6 == nil {
```
hebelsan: Just for clarification: it's not an issue if both `SubnetPoolID` and `Networks.IPv6` are set right?
I assume in this case it will allocate the given subnets from the subnetPool that is  provided.

pkg/controller/infrastructure/infraflow/utils.go
```diff
@@ -196,6 +197,10 @@ func (fctx *FlowContext) defaultSubnetName() string {
 	return fctx.infra.Namespace
 }
 
+func (fctx *FlowContext) defaultSubnetIPv6Name() string {
+	return fctx.infra.Namespace + "-ipv6"
```
hebelsan, line 201: The shoot ID might be more suitable then the `Namespace` because it could be that the same shoot name exists on dev and canary and it uses the same credentials.

pkg/controller/infrastructure/infraflow/utils.go
```diff
@@ -207,3 +212,14 @@ func (fctx *FlowContext) defaultSharedNetworkName() string {
 func (fctx *FlowContext) workersCIDR() string {
 	return fctx.config.WorkersCIDR()
 }
+
+// isDualStack returns true if the cluster is configured for dual-stack networking
+// or has migrated from dual-stack to single-stack (indicated by having 2 node CIDRs).
+func (fctx *FlowContext) isDualStack() bool {
```
hebelsan: We can use the g/g function for that if you rebase.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -70,10 +77,24 @@ func (fctx *FlowContext) buildReconcileGraph() *flow.Graph {
...
 		shared.Timeout(defaultTimeout), shared.Dependencies(ensureRouter, ensureSubnet))
 
+	_ = fctx.AddTask(g, "ensure IPv6 router interface",
+		fctx.ensureRouterInterfaceIPv6,
+		shared.Timeout(defaultTimeout), shared.Dependencies(ensureRouter, ensureSubnetIPv6))
+
+	_ = fctx.AddTask(g, "ensure IPv6 CIDR services", fctx.ensureIPv6CIDRs,
+		shared.Timeout(defaultTimeout),
+		shared.Dependencies(ensureSubnetIPv6),
+		shared.DoIf(fctx.isDualStack()),
```
hebelsan, line 92: The tasks `ensure IPv6 subnet` and `ensure IPv6 router interface` perform the `fctx.isDualStack()` check internally, while here it is applied at the task definition level using `shared.DoIf(fctx.isDualStack())`.
Please align the implementation to use a consistent approach.

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -42,7 +45,11 @@ func (fctx *FlowContext) Reconcile(ctx context.Context) error {
 
 	state := fctx.computeInfrastructureState()
 	status := fctx.computeInfrastructureStatus()
-	return PatchProviderStatusAndState(ctx, fctx.client, fctx.infra, status, state)
+	nodeCIDR := fctx.state.Get(IdentifierNodeSubnetIPv6CIDR)
+	podCIDR := fctx.state.Get(IdentifierPodSubnetIPv6CIDR)
+	svcCIDR := fctx.state.Get(IdentifierServiceSubnetIPv6CIDR)
+	return PatchProviderStatusAndState(ctx, fctx.client, fctx.infra, fctx.shootNetworking, status, state, nodeCIDR, podCIDR, svcCIDR)
+	//return PatchProviderStatusAndState(ctx, fctx.client, fctx.infra, fctx.shootNetworking, status, state, nil, nil, nil)
```
hebelsan: I guess this comment line can be removed.

pkg/controller/infrastructure/infraflow/context.go
```diff
@@ -135,28 +150,57 @@ func NewFlowContext(opts Opts) (*FlowContext, error) {
...
+
+	if nodesSubnetIPv6CIDR != nil {
+		infra.Status.Networking.Nodes = append(infra.Status.Networking.Nodes, *nodesSubnetIPv6CIDR)
+	}
+	if podSubnetIPv6CIDR != nil {
+		infra.Status.Networking.Pods = append(infra.Status.Networking.Pods, *podSubnetIPv6CIDR)
+	}
+	if servicesSubnetIPv6CIDR != nil {
+		infra.Status.Networking.Services = append(infra.Status.Networking.Services, *servicesSubnetIPv6CIDR)
+	}
```
hebelsan: What do you think about moving this computation into `fctx.computeInfrastructureStatus()`?
It could help keep the parameters of `PatchProviderStatusAndState()` clean and lightweight.

hebelsan (commented):
Thanks for the contribution! 
I’ve left a few comments.

charts/internal/machineclass/templates/machineclass.yaml
axel7born: I removed it. However, while testing I found a bug in [machine-controller-manager-provider-openstack](https://github.com/gardener/machine-controller-manager-provider-openstack).
I opened a PR to fix it: https://github.com/gardener/machine-controller-manager-provider-openstack/pull/373

axel7born:
@hebelsan I think I addressed your comments. 
However, we need to wait for 
https://github.com/gardener/machine-controller-manager-provider-openstack/pull/373
/hold

hebelsan:
MCM got updated with https://github.com/gardener/gardener-extension-provider-openstack/pull/1313

pkg/controller/infrastructure/infraflow/delete.go
```diff
@@ -186,6 +222,24 @@ func (fctx *FlowContext) recoverSubnetID(ctx context.Context) error {
...
 }
 
+func (fctx *FlowContext) recoverSubnetIPv6ID(ctx context.Context) error {
+	for _, suffix := range []string{"-ipv6-pod", "-ipv6-svc", "-ipv6"} {
+		identifier := getSubnetIdentifierBySuffix(suffix)
+		if fctx.state.Get(identifier) != nil {
+			return nil
+		}
+
+		subnet, err := fctx.findExistingSubnetIPv6(ctx, fctx.defaultSubnetIPv6Name()+suffix)
```
hebelsan: The suffix list ["-ipv6-pod", "-ipv6-svc", "-ipv6"] is iterated, which means findExistingSubnetIPv6 is called with defaultSubnetIPv6Name() + suffix = "{ns}-ipv6" + "-ipv6-pod" = "{ns}-ipv6-ipv6-pod" — double "-ipv6". I think that will never match.

pkg/controller/infrastructure/infraflow/delete.go
```diff
@@ -186,6 +222,24 @@ func (fctx *FlowContext) recoverSubnetID(ctx context.Context) error {
 	return nil
 }
 
+func (fctx *FlowContext) recoverSubnetIPv6ID(ctx context.Context) error {
+	for _, suffix := range []string{"-ipv6-pod", "-ipv6-svc", "-ipv6"} {
+		identifier := getSubnetIdentifierBySuffix(suffix)
+		if fctx.state.Get(identifier) != nil {
+			return nil
```
hebelsan: Is the early return ok here? 
Because if only IdentifierSubnetIPv6 is already in state, recovery of pod/service subnets is skipped entirely.

pkg/controller/infrastructure/infraflow/delete.go
```diff
@@ -147,6 +169,20 @@ func (fctx *FlowContext) deleteSubnet(ctx context.Context) error {
 	return nil
 }
 
+func (fctx *FlowContext) deleteSubnetIPv6(ctx context.Context) error {
```
hebelsan, line 172: Where is the deletion of pod (IdentifierSubnetIPv6Pod) and service (IdentifierSubnetIPv6Svc) subnets?

pkg/controller/infrastructure/infraflow/context.go
```diff
@@ -214,12 +265,17 @@ func (fctx *FlowContext) computeInfrastructureStatus() *openstackv1alpha1.Infras
...
-				ID:      *v,
-			},
-		}
+		status.Networks.Subnets = append(status.Networks.Subnets, openstackv1alpha1.Subnet{
+			Purpose: openstackv1alpha1.PurposeNodes,
+			ID:      *v,
+		})
+	}
+
+	if v := fctx.state.Get(IdentifierSubnetIPv6); v != nil {
```
hebelsan, line 274: what about adding `IdentifierSubnetIPv6Pod` and `IdentifierSubnetIPv6Svc`?

hebelsan (commented):
I've got some more comments.

pkg/controller/infrastructure/infraflow/delete.go
axel7born: Fixed.

pkg/controller/infrastructure/infraflow/delete.go
axel7born: Fixed.

pkg/controller/infrastructure/infraflow/delete.go
axel7born: Good catch. Added the deletion of svc and pod subnet.

pkg/controller/infrastructure/infraflow/context.go
axel7born: I added them to the state and made sure only the nodes subnet ip is used for machine creation.

pkg/apis/openstack/types_infrastructure.go
```diff
@@ -18,6 +18,9 @@ type InfrastructureConfig struct {
 	// FloatingPoolSubnetName contains the fixed name of subnet or matching name pattern for subnet
 	// in the Floating IP Pool where the router should be attached to.
 	FloatingPoolSubnetName *string
+	// SubnetPoolID is the ID of the subnet pool to use for IPv6 subnet allocation.
+	// If not specified, IPv6 subnets will be allocated from the default subnet pool.
+	SubnetPoolID *string
```
hebelsan: If this subnetPoolID is only used for IPv6, wouldn't it make sense to rename it to IPv6SubnetPoolID or put it to the IPv6 config?
Or might this field in future also be used for IPv4 cluster?

docs/usage/usage.md
```diff
@@ -87,6 +87,47 @@ Apart from the router and the worker subnet the OpenStack extension will also cr
...
+```yaml
+apiVersion: openstack.provider.extensions.gardener.cloud/v1alpha1
+kind: InfrastructureConfig
+floatingPoolName: MY-FLOATING-POOL
+networks:
+  workers: 10.250.0.0/19
+  ipv6:
+    nodeCIDR: 2001:db8:1::/64      # IPv6 CIDR for worker nodes
+    podCIDR: 2001:db8:2::/64        # IPv6 CIDR for pods
+    serviceCIDR: 2001:db8:3::/112   # IPv6 CIDR for services
```
hebelsan, line 127: It seems like the validation for ipv6 is missing.
we should validate immutability, valid IPv6 cidr syntax, and no overlapping between them.

pkg/controller/worker/machines.go
```diff
@@ -201,7 +201,15 @@ func (w *WorkerDelegate) generateMachineConfig(ctx context.Context) error {
 				},
 			}
 
-			machineClassSpec["subnetID"] = subnet.ID
+			// Collect only node subnet IDs; pod/service subnets must not be attached to machines.
+			var subnetIDs []string
+			for _, subnet := range subnets {
+				if subnet.Purpose == api.PurposeNodes {
```
hebelsan: The check for PurposeNodes here seems redundant, since `FindSubnetsByPurpose` already filters.

hebelsan (commented):
In the file _valuesprovider.go_ line 513 `FindSubnetByPurpose` returns the first PurposeNodes subnet.
The subnetID is passed to the cloud-provider-openstack config chart.
This might break now with dual-stack where both IPv4 and IPv6 node subnets have PurposeNodes.
Instead of relying on the ordering in the status we should make sure that it always uses the IPv4 Subnet.

pkg/controller/infrastructure/infraflow/context.go
```diff
@@ -135,12 +150,13 @@ func NewFlowContext(opts Opts) (*FlowContext, error) {
...
 		client:                 opts.Client,
 		openstackClientFactory: opts.ClientFactory,
+		shootNetworking:        opts.Cluster.Shoot.Spec.Networking,
 	}
 	return flowContext, nil
 }
 
 func (fctx *FlowContext) persistState(ctx context.Context) error {
-	return PatchProviderStatusAndState(ctx, fctx.client, fctx.infra, nil, fctx.computeInfrastructureState())
+	return PatchProviderStatusAndState(ctx, fctx.client, fctx.infra, nil, fctx.computeInfrastructureState(), nil)
```
hebelsan: `nil` is unconditionally assigned to infra.Status.Networking. 
If reconciliation partially succeeds and then a step fails, the intermediate persistState call will wipe the networking status.
That would be a regression for both single-stack and dual-stack shoots. 
```suggestion
	return PatchProviderStatusAndState(ctx, fctx.client, fctx.infra, nil, fctx.computeInfrastructureState(), fctx.computeInfrastructureNetworkingStatus())
```

hebelsan (commented):
Thanks for the fast iterations and this contribution in general!
/lgtm
/approve

PR #1269, opened 2026-02-11
Update usage doc for supporting virtualCapacities on nodeTemplates

aaronfern:

/area auto-scaling
/kind enhancement
/platform openstack

This PR is a follow up of #1266 to update the usage doc

/cc @hebelsan 

**Release note**:

```other operator
NONE
```


aaronfern:
Right, I did not see that PR before

I can modify this PR to drop the unit test and have only the change is the usage doc

PR #1317, opened 2026-04-15
Feature/cloud profile storage type

### LINKED ISSUE #1316: Add support for kvm machines

/area usability
/kind enhancement
/platform openstack

SAP Cloud now supports KVM-based machines.
These machines require a specific disk type, “premium,” for the boot disk.
Currently, if no disk type is specified in the worker section of the Shoot YAML, the default disk type defined in the corresponding Cloud Provider (CC) project is used. In most cases, this default is set to “vmware.”
To support KVM machines properly, we need to update the Cloud Profile configuration:

- The `storage.type` in the machineTypes section must be set to “premium” for KVM-based machine types.
For example:
```yaml
spec:
  machineTypes:
  - architecture: amd64
    cpu: "4"
    gpu: "0"
    memory: 8Gi
    name: g_k_c4_m8_v2
    storage:
      class: standard
      size: "64"
      type: premium
    usable: true
```

Additionally, the OpenStack extension should:
 
- Use the `storage.type` defined in the Cloud Profile as the default value when creating machines, if no explicit value is provided in the Shoot specification.

**Why is this needed**:
To natively support kvm machines.

hebelsan:

/area quality
/kind enhancement
/platform openstack

This PR adjusts the default values for the machineclass values `rootDiskType` and `rootDiskSize`.
These will now be derived from the cloudProfile machineType spec which looks the following:
```yaml
spec:
  machineTypes:
  - architecture: amd64
    cpu: "4"
    gpu: "0"
    memory: 8Gi
    name: g_k_c4_m8_v2
    storage:
      class: standard
      size: "64"
      type: premium
    usable: true
```
If the storage section of the machineType is not set or the type is set to default the machinclass fields for `rootDiskType` and `rootDiskSize` will be empty so the openstack defaults will be used.
It's still possible to overwrite the default disk type and size by setting the shoot fields.

Fixes https://github.com/gardener/gardener-extension-provider-openstack/issues/1316

**Release note**:

```breaking operator
Behavioral change: rootDiskType and rootDiskSize now derived from cloud profile machine type storage.
Previously, if a worker pool did not explicitly configure volume.type / volume.size, the generated MachineClass omitted rootDiskType and rootDiskSize entirely, leaving the choice to OpenStack's infrastructure defaults.
With this change, if the cloud profile's MachineType.Storage defines a type (other than "default") and optionally a storageSize, those values are now used as defaults in the generated MachineClass.
Impact: Shoots using machine types whose cloud profile entry has a non-"default" Storage.Type will now have an explicit rootDiskType (and potentially rootDiskSize) set in their MachineClass. If this differs from the OpenStack infrastructure's previous default, the root disk type of newly created nodes will change.
Action required: Review the Storage field on machine types in affected cloud profiles. If the Storage.Type was set for informational purposes only and should not influence disk provisioning, set it to "default" to preserve the previous behavior.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,8 @@ TODO
 # GitGuardian
 .cache_ggshield
 .gitguardian.yaml
+
+# AI coding agents
+# Claude
+.claude/
+/skills-lock.json
--- a/pkg/controller/worker/machines.go
+++ b/pkg/controller/worker/machines.go
@@ -210,6 +210,20 @@ func (w *WorkerDelegate) generateMachineConfig(ctx context.Context) error {
 			// specifying the volume type requires a custom volume size to be specified too.
 			if pool.Volume != nil && pool.Volume.Type != nil {
 				machineClassSpec["rootDiskType"] = *pool.Volume.Type
+			} else if machineTypeFromCloudProfile.Storage != nil &&
+				machineTypeFromCloudProfile.Storage.Type != "" &&
+				machineTypeFromCloudProfile.Storage.Type != "default" {
+				// Use the storage type from the cloud profile as the default if not explicitly set in the shoot spec.
+				// This is required e.g. for KVM machines that need a "premium" disk type for boot disks.
+				machineClassSpec["rootDiskType"] = machineTypeFromCloudProfile.Storage.Type
+				if machineTypeFromCloudProfile.Storage.StorageSize != nil {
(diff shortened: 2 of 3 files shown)


AndreasBurger (commented):
/approve

AndreasBurger (approved):
since auto-merge is blocked by milestone anyways and this l's gtm I can actually properly approve this.

PR #1326, opened 2026-04-28
`feature:` Support OpenStack subnet pools for automatic CIDR allocation

### LINKED ISSUE #1324: Support OpenStack subnet pools for automatic CIDR allocation in InfrastructureConfig
/area networking
/kind enhancement
/platform openstack

Support for OpenStack subnet pools in the `InfrastructureConfig`. Instead of requiring a static CIDR in `networks.workers`, allow specifying a subnet pool name and prefix length. OpenStack Neutron will then automatically allocate a free CIDR from the pool.

Example configuration:

```yaml
kind: InfrastructureConfig
floatingPoolName: my-floating-pool
networks:
  id: <existing-network-id>
  subnetPool:
    name: "my-subnet-pool"
    prefixLength: 24
  router:
    id: <existing-router-id>
```

This would translate to the equivalent OpenStack API call:

```bash
openstack subnet create --network <net> \
  --subnet-pool "my-subnet-pool" \
  --prefix-length 24 \
  my-subnet
```

The allocated CIDR should be reported back via the `InfrastructureStatus` and used by Gardener for VPN and node network configuration.

**Why is this needed**:

Many OpenStack environments, especially in enterprise and government infrastructure, use subnet pools to manage IP address allocation centrally. Administrators assign a subnet pool to a project and expect all subnets to be allocated from that pool automatically.

Currently, the `InfrastructureConfig` only accepts a static CIDR via `networks.workers`. This requires users to manually find a free CIDR range from their pool before creating a shoot, which is error prone and conflicts with the centralized IP management approach that subnet pools are designed for.

Supporting subnet pools would allow Gardener to integrate naturally with existing OpenStack IP address management workflows without requiring manual CIDR coordination.

hebelsan:

/area usability
/kind enhancement
/platform openstack

Adds the possibility to use a subnet pools instead of predefined CIDRs.

Fixes https://github.com/gardener/gardener-extension-provider-openstack/issues/1324

**Release note**:

```other operator
support OpenStack subnet pools for automatic CIDR allocation
```


docs/usage/usage.md
```diff
@@ -82,6 +82,20 @@ The `networks.workers` section describes the CIDR for a subnet that is used for
...
+Instead of specifying an explicit CIDR via `networks.workers`, you can use `networks.subnetPool` to let OpenStack automatically allocate a subnet CIDR from a pre-configured subnet pool:
+
+```yaml
+apiVersion: openstack.provider.extensions.gardener.cloud/v1alpha1
+kind: InfrastructureConfig
+floatingPoolName: MY-FLOATING-POOL
+networks:
+  subnetPool:
+    id: MY-SUBNET-POOL-ID
+    prefixLength: 24  # request a /24 subnet from the pool
```
kon-angelo, line 94: 24 is by default a single node. I would say at least make it 16 to be more realistic

pkg/apis/openstack/validation/infrastructure.go
```diff
@@ -136,8 +162,12 @@ func ValidateInfrastructureConfigUpdate(oldConfig, newConfig *api.Infrastructure
 	// networks.ipv6 is immutable and checked explicitly below for a precise error field.
 	newNetworks.IPv6 = nil
 	oldNetworks.IPv6 = nil
+	// networks.subnetPool is immutable and checked explicitly below for a precise error field.
+	newNetworks.SubnetPool = nil
+	oldNetworks.SubnetPool = nil
```
kon-angelo: why do we do that exactly when we want both fields to be immutable anyway ?

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -974,10 +974,18 @@ func (vp *valuesProvider) addCSIManilaValues(
...
 		shareNetworkID = infraStatus.Networks.ShareNetwork.ID
 	}
+	shareClient := infraConfig.WorkersCIDR()
+	if shareClient == "" {
+		// When a subnet pool is used, the workers CIDR is not statically configured but allocated dynamically.
+		// In that case, the allocated CIDR is stored in the nodes subnet status.
+		if subnet, err := helper.FindSubnetByPurpose(infraStatus.Networks.Subnets, api.PurposeNodes); err == nil {
+			shareClient = subnet.CIDR
+		}
+	}
```
kon-angelo: let's make it to always use the status. I don't see a reason for the split code path here.

pkg/controller/infrastructure/infraflow/access/networking_access.go
```diff
@@ -431,6 +431,10 @@ func (a *networkingAccess) CreateSubnet(ctx context.Context, desired *subnets.Su
 		opts.SubnetPoolID = desired.SubnetPoolID
 	}
 
+	if prefixlen != nil {
+		opts.Prefixlen = *prefixlen
```
kon-angelo: if the opts.Prefixlen parameter is not a pointer why not always use the ptr.Deref ?

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -320,13 +319,32 @@ func (fctx *FlowContext) ensureSubnet(ctx context.Context) error {
...
+		}
 	} else {
 		log.Info("creating...")
-		created, err := fctx.access.CreateSubnet(ctx, desired)
+		var subnetPoolID string
+		var prefixlen *int
+		if fctx.config.Networks.SubnetPool != nil {
+			subnetPoolID = fctx.config.Networks.SubnetPool.ID
+			pl := fctx.config.Networks.SubnetPool.PrefixLength
+			prefixlen = &pl
```
kon-angelo: hb
```
pl:=ptr.To(fctx.config.Networks.SubnetPool.PrefixLength)
```

pkg/openstack/client/networking.go
```diff
@@ -284,6 +285,25 @@ func (c *NetworkingClient) DeleteSubnet(ctx context.Context, subnetID string) er
...
+	if err != nil {
+		return nil, err
+	}
+	return subnetpools.ExtractSubnetPools(page)
+}
+
+// CreateSubnetPool creates a subnet pool
+func (c *NetworkingClient) CreateSubnetPool(ctx context.Context, createOpts subnetpools.CreateOpts) (*subnetpools.SubnetPool, error) {
+	return subnetpools.Create(ctx, c.client, createOpts).Extract()
+}
```
kon-angelo, line 300: am I right that this is this only used for testing ?

docs/usage/usage.md
hebelsan: What exactly do you mean 24 is a single node?
This is about a /24 subnet prefix length (CIDR notation). 
It means the subnet is 256 IPs large.

pkg/apis/openstack/validation/infrastructure.go
hebelsan: `ValidateImmutableField(newNetworks, oldNetworks, ...)` does a broad diff of the entire networks struct.
If SubnetPool were left in, a change to it would produce a generic error like _"networks: value is immutable"_, which is not very helpful.
That's why we nil them and validate separately.

pkg/openstack/client/networking.go
hebelsan: Yes. Should I create helper functions for creating a deleting subnet pools in the test files instead?

pkg/apis/openstack/validation/infrastructure.go
kon-angelo: then nit: maybe instead of doing "exception-based" lines of code, we write this in a more finegrained but easier to follow format ?

I would even accept a TODO comment for next development cycle.

pkg/openstack/client/networking.go
kon-angelo: This is a tricky topic. You want minimal dependencies between your test code and production but maintaining 2 different set of clients is troublesome. Not sure if a comment would make any difference if not maintained properly. No action needed

docs/usage/usage.md
kon-angelo: Sorry I was somehow confused with pod network for which a /24 would be small. Please disregard

pkg/apis/openstack/validation/infrastructure.go
hebelsan: I've changed the code to now explicitly validate all `infraConfig.Networks` fields.

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -952,10 +948,14 @@ func (vp *valuesProvider) addCSIManilaValues(
...
 	}
+	nodesSubnet, err := helper.FindSubnetByPurpose(infraStatus.Networks.Subnets, api.PurposeNodes)
+	if err != nil {
+		return fmt.Errorf("could not find nodes subnet in infrastructure status: %w", err)
+	}
 	values["openstack"] = map[string]interface{}{
 		"availabilityZones":           vp.getAllWorkerPoolsZones(cluster),
 		"shareNetworkID":              shareNetworkID,
-		"shareClient":                 infraConfig.WorkersCIDR(),
+		"shareClient":                 nodesSubnet.CIDR,
```
kon-angelo, line 958: shareClient now reads from nodesSubnet.CIDR — but CIDR is only populated for subnet-pool-allocated subnets.

pkg/controller/infrastructure/infraflow/context.go
```diff
@@ -265,10 +272,14 @@ func (fctx *FlowContext) computeInfrastructureStatus() *openstackv1alpha1.Infras
 	}
 
 	if v := fctx.state.Get(IdentifierSubnet); v != nil {
-		status.Networks.Subnets = append(status.Networks.Subnets, openstackv1alpha1.Subnet{
+		subnet := openstackv1alpha1.Subnet{
 			Purpose: openstackv1alpha1.PurposeNodes,
 			ID:      *v,
-		})
+		}
+		if cidr := fctx.state.Get(IdentifierWorkersCIDR); cidr != nil {
```
kon-angelo: Does this have to be conditional (also see the comment about `valuesprovider`)

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -320,13 +319,31 @@ func (fctx *FlowContext) ensureSubnet(ctx context.Context) error {
...
+		if fctx.config.Networks.SubnetPool != nil {
+			fctx.state.Set(IdentifierWorkersCIDR, current.CIDR)
+		}
 	} else {
 		log.Info("creating...")
-		created, err := fctx.access.CreateSubnet(ctx, desired)
+		var subnetPoolID string
+		var prefixlen *int
+		if fctx.config.Networks.SubnetPool != nil {
+			subnetPoolID = fctx.config.Networks.SubnetPool.ID
```
kon-angelo: why not set that directly to desired?

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -320,13 +319,31 @@ func (fctx *FlowContext) ensureSubnet(ctx context.Context) error {
...
+			prefixlen = ptr.To(fctx.config.Networks.SubnetPool.PrefixLength)
+			desired.CIDR = "" // let the pool allocate
```
kon-angelo: Maybe instead of emptying the value, we could assign it as needed (in our usual happy path). Essentially inverse this for better readability. WDYT?

pkg/apis/openstack/v1alpha1/types_infrastructure.go
```diff
@@ -157,6 +170,9 @@ type Subnet struct {
 	Purpose Purpose `json:"purpose"`
 	// ID is the subnet id.
 	ID string `json:"id"`
+	// CIDR is the CIDR of the subnet. This is only set for subnets that were allocated from a subnet pool.
+	// +optional
+	CIDR string `json:"cidr,omitempty"`
```
kon-angelo: Why only conditionally populate the field ? I wouldn't make this field the marker for whether pools or not are used if that's the main idea for introducing this.

pkg/apis/openstack/v1alpha1/types_infrastructure.go
```diff
@@ -33,7 +33,12 @@ type Networks struct {
 	// Deprecated: use `workers` instead.
 	Worker string `json:"worker"`
 	// Workers is a CIDRs of a worker subnet (private) to create (used for the VMs).
+	// Mutually exclusive with SubnetPool.
 	Workers string `json:"workers"`
```
kon-angelo: Is that not now optional ?

hebelsan: yes, worker (deprecated) and workers are now optional.
I've added the missing ` // +optional` but given that omitempty on a plain string already handles the JSON serialization correctly, the pointer change would be a refactor with no benefit in my opinion.

pkg/controller/controlplane/valuesprovider.go
hebelsan: good catch, changed it so we always populate the CIDR field.

pkg/apis/openstack/v1alpha1/types_infrastructure.go
hebelsan: Good point.
Switched to not optional and always populate it.

pkg/controller/infrastructure/infraflow/reconcile.go
hebelsan: Changed it to set it directly.

pkg/controller/infrastructure/infraflow/reconcile.go
hebelsan: Sounds good, adjusted the logic to assign the values as needed.

pkg/controller/infrastructure/infraflow/context.go
hebelsan: fixed it to always set the cidr.

PR #1327, opened 2026-04-29
`feature:` support for additional security groups in worker configuration

### LINKED ISSUE #1031: Add support for existing SecurityGroups for MachineDeployment via WorkerConfig API

/area TODO
/kind enhancement
/platform openstack

Add support for specifying existing SecurityGroups for each MachineDeployment (worker group) via the WorkerConfig API.

**Why is this needed**:
Currently, the SecurityGroup of a cluster is defined using Terraform inside [./pkg/internal/infrastructure/templates](https://github.com/gardener/gardener-extension-provider-openstack/tree/master/pkg/internal/infrastructure/templates).
In some cases, Kubernetes (shoot) clusters are designed to run multiple applications (e.g., Kafka, MongoDB, Redis,...) in separate worker groups and expose each service to clients using L2 LoadBalancers. Each worker group requires a specific SecurityGroup to function securely and correctly.

Since the MachineControllerManager (MCM) already supports multiple SecurityGroups per MachineDeployment, I would like to extend the WorkerConfig API to allow referencing existing SecurityGroups. This will enable more flexible and secure network configurations tailored to application-specific needs.

hebelsan:

/area quality
/kind enhancement
/platform openstack

This PR adds support for additional security groups in worker configuration.

Fixes https://github.com/gardener/gardener-extension-provider-openstack/issues/1031

**Release note**:

```feature operator
support for additional security groups in worker configuration
```

### DIFF
--- a/pkg/apis/openstack/validation/worker_test.go
+++ b/pkg/apis/openstack/validation/worker_test.go
@@ -328,6 +328,15 @@ var _ = Describe("ValidateWorkerConfig", func() {
 				})),
 			))
 		})
+
+		It("should return an error for duplicate security group names", func() {
+			Expect(ValidateAdditionalSecurityGroups([]string{"sg-1", "sg-2", "sg-1"}, fldPath.Child("additionalSecurityGroups"))).To(ConsistOf(
+				PointTo(MatchFields(IgnoreExtras, Fields{
+					"Type":  Equal(field.ErrorTypeInvalid),
+					"Field": Equal("config.additionalSecurityGroups[2]"),
+				})),
+			))
+		})
 	})
 
 	Describe("#ValidateNodeTemplate", func() {
--- a/pkg/apis/openstack/validation/worker_test.go
+++ b/pkg/apis/openstack/validation/worker_test.go
@@ -309,6 +309,27 @@ var _ = Describe("ValidateWorkerConfig", func() {
 		})
 	})
 
+	Describe("#ValidateAdditionalSecurityGroups", func() {
+		var fldPath = field.NewPath("config")
+
+		It("should return no errors for an empty list", func() {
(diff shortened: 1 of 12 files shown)


pkg/apis/openstack/validation/worker_test.go
```diff
@@ -309,6 +309,36 @@ var _ = Describe("ValidateWorkerConfig", func() {
...
+	Describe("#ValidateAdditionalSecurityGroups", func() {
+		var fldPath = field.NewPath("config")
+
+		It("should return no errors for an empty list", func() {
+			Expect(ValidateAdditionalSecurityGroups(nil, fldPath.Child("additionalSecurityGroups"))).To(BeEmpty())
+		})
+
+		It("should return no errors for valid names", func() {
+			Expect(ValidateAdditionalSecurityGroups([]string{"sg-1", "my-security-group"}, fldPath.Child("additionalSecurityGroups"))).To(BeEmpty())
+		})
```
AndreasBurger, line 321: A valid name right now is any name that is not empty. Since we already validate that empty names do not pass validation, this test does not provide any additional benefit.

If we have some canonical description of a valid sg-name, we could use that for an actual validation instead, but I'm fine with it as is. The introduced names should be valid by virtue of stemming from previously created sg-groups on the openstack instance.

AndreasBurger (commented):
/approve

lgtm. One comment, if you want to change it. Otherwise let me know and we'll merge it.

pkg/apis/openstack/validation/worker_test.go
hebelsan: I'd leave it for now, since it's the only positive test for additionalSecurityGroups.

AndreasBurger: That statement basically shows that it's worse, as this suggests something being tested that is, in fact, not.

On the other hand, the only one that this hurts is the testrunner, so fair enough.

AndreasBurger (approved):

PR #1333, opened 2026-05-06
[GEP-33] Allow mixed format Capability CloudProfiles

Roncossek:
/area control-plane
/kind enhancement
/platform openstack

Port of [gardener-extension-provider-aws#1645](https://github.com/gardener/gardener-extension-provider-aws/pull/1645) to OpenStack: allows old format (regions with architecture) and new format (capabilityFlavors) to coexist per image version within CloudProfiles and NamespacedCloudProfiles.

- Added `CloudProfile` and `NamespacedCloudProfile` mutators to populate `capabilityFlavors` from providerConfig
- Updated validation to accept both formats per version; skip validation for expirationDate-only parent version overrides
- Refactored `FindImageInCloudProfile` to handle mixed formats; deleted `cloudprofiletransformer.go`
- Updated bastion to use shared image lookup

Fixes #

- The extensionswebhook handler routes by GVK only — a single combined ´nscpf´ mutator handles both spec and status subresource mutations.

**Release note**:
```feature operator
Allow mixed format machine images in CloudProfiles: old format (regions with architecture) and new format (capabilityFlavors) can coexist per image version.
```


pkg/admission/validator/namespacedcloudprofile.go
```diff
@@ -76,10 +75,6 @@ func (p *namespacedCloudProfile) Validate(ctx context.Context, newObj, _ client.
 		allErrs = append(allErrs, err.(*field.Error))
```
hebelsan: I know this wasn't introduced by this PR but the error casting here looks suspicious.
Wouldn't it make sense to change the return type of `validateMachineImagesOnlyInNamespacedCloudProfile` to `field.ErrorList`?

Roncossek: Sounds about right. I will investigate why that decision was made in the past. It a change is possible i would suggest to create the same PR on all provider repos at once. I would take that over.

pkg/controller/worker/machine_images.go
hebelsan, line 125: Why was
```
Image:        machineImage.Image,
```
not added here?

Roncossek: Good catch. I added the Image field and adjusted the tests.

pkg/admission/validator/namespacedcloudprofile.go
Roncossek: implemented the type changed in this PR 👍

PR #1334, opened 2026-05-07
Remove controllers flag from deployment configuration

hebelsan:

/area quality
/kind enhancement
/platform openstack

In the context of gardener/gardener#14628, we need to enable the BackupEntry Controller in provider extensions. Currently, it's disabled, so we have to introduce a feature gate (with gardener/gardener#14749), otherwise this won't work and the BackupEntry never gets reconciled.

Fixes #

**Release note**:

```other operator
Remove controllers flag from deployment configuration
```


hebelsan:
While the reconciliations are scoped, the controllers still register a watch on all resources.
This fails either because the CRDs don't even exist in the runtime cluster or because the service account does not have permissions.
The controller pods are continuously crashing in the runtime cluster.

PR #1341, opened 2026-05-19
Bring extension main Pod back under VPA control for CPU

voelzmo:

/area control-plane
/kind bug
/platform openstack

PR https://github.com/gardener/gardener-extension-provider-openstack/pull/1215 accidentally removed CPU from the VPA   `controlledResources` for the main extension pod, leaving it only managing memory. In practice, the default CPU requests are too low   compared to real-world usage, so this restores CPU autoscaling by adding `cpu` back to `controlledResources`.

Additionally, the wildcard `containerName: '*'` policy is replaced with an explicit policy for the  `gardener-extension-provider-openstack` container plus a catch-all `'*'` policy with `mode: "Off"`. This follows the Gardener  project's VPA best practices ([defining a VPA resource](https://gardener.cloud/docs/guides/applications/shoot-pod-autoscaling-best-practices/#defining-a-vpa-resource)), which recommend explicitly disabling VPA for all containers not under active control to prevent unintended scaling of injected sidecars.

Fixes #

**Release note**:
```bugfix operator
VPA for the extension pod now also controls CPU requests in addition to memory. The VPA container policy also explicitly disables autoscaling for any non-main containers following Gardener best practices.
```


wpross:
What about the `gardener-extension-admission-openstack` image?

voelzmo:
> What about the `gardener-extension-admission-openstack` image?

I don't think we need to change anything for this Deployment at the moment

wpross:
> > What about the `gardener-extension-admission-openstack` image?
> 
> I don't think we need to change anything for this Deployment at the moment

Because of the lower cpu usage profile of the admission image compared to the provider image?

voelzmo:
> > > What about the `gardener-extension-admission-openstack` image?
> > 
> > 
> > I don't think we need to change anything for this Deployment at the moment
> 
> Because of the lower cpu usage profile of the admission image compared to the provider image?

Yes, we see it running with 30-50mCores CPU and it has 50mCores requests, so this seems to fit quite nicely.

wpross (approved):
/lgtm

PR #1347, opened 2026-05-26
fix: filter DNS servers by IP family in reconcile logic

### LINKED ISSUE #1346: dual-stack subnet creation fails due to IPv4 nameserver

/area robustness
/kind bug
/platform openstack

**What happened**:

When trying to create a dual-stack shoot, the IPv6 subnet creation fails due to dnsServers in cloudprofile containing IPv4 addresses.

Filter DNS server items to the correct address families each.

Create a shoot with IPv4 and IPv6 `ipFamilies` and a subnetPoolID given.

**Environment**:

- Gardener version (if relevant): v1.142.3
- Extension version: v1.55.0
- Kubernetes version (use `kubectl version`): 1.33.12
- Cloud provider or hardware configuration: vanilla-ish openstack (ScaleUp Open Cloud)
- Others: -

hebelsan:

/area quality
/kind enhancement
/platform openstack

When creating a dual-stack shoot cluster (IPv4 + IPv6), the IPv6 subnet creation fails because the dnsServers in the cloud profile contain only IPv4 addresses. OpenStack rejects IPv4 DNS servers when configuring an IPv6 subnet.

Fixes https://github.com/gardener/gardener-extension-provider-openstack/issues/1346

**Release note**:

```bugfix operator
Filter subnet DNS servers from CloudProfile by IP family in reconcile logic
```

### DIFF
--- a/pkg/controller/infrastructure/infraflow/reconcile.go
+++ b/pkg/controller/infrastructure/infraflow/reconcile.go
@@ -306,7 +306,7 @@ func (fctx *FlowContext) ensureSubnet(ctx context.Context) error {
 		Name:           fctx.defaultSubnetName(),
 		NetworkID:      networkID,
 		IPVersion:      4,
-		DNSNameservers: fctx.cloudProfileConfig.DNSServers,
+		DNSNameservers: filterDNSServersByIPFamily(fctx.cloudProfileConfig.DNSServers, gardencorev1beta1.IPFamilyIPv4),
 	}
 	if fctx.config.Networks.SubnetPool != nil {
 		desired.SubnetPoolID = fctx.config.Networks.SubnetPool.ID
@@ -378,7 +378,7 @@ func (fctx *FlowContext) ensureSubnetIPv6(ctx context.Context) error {
 			NetworkID:       networkID,
 			CIDR:            nodeCIDR,
 			IPVersion:       6,
-			DNSNameservers:  fctx.cloudProfileConfig.DNSServers,
+			DNSNameservers:  filterDNSServersByIPFamily(fctx.cloudProfileConfig.DNSServers, gardencorev1beta1.IPFamilyIPv6),
 			IPv6RAMode:      "slaac",
 			IPv6AddressMode: "slaac",
 			SubnetPoolID:    subnetPoolID,
@@ -388,7 +388,7 @@ func (fctx *FlowContext) ensureSubnetIPv6(ctx context.Context) error {
 			NetworkID:      networkID,
 			CIDR:           podCIDR,
 			IPVersion:      6,
-			DNSNameservers: fctx.cloudProfileConfig.DNSServers,
+			DNSNameservers: filterDNSServersByIPFamily(fctx.cloudProfileConfig.DNSServers, gardencorev1beta1.IPFamilyIPv6),
 			SubnetPoolID:   subnetPoolID,
 		},
 		{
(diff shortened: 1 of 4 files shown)


pkg/controller/infrastructure/infraflow/utils.go
```diff
@@ -62,6 +62,20 @@ func sliceToPtr[T any](slice []T) []*T {
 	return res
 }
 
+// filterDNSServersByIPFamily returns only the DNS servers matching the given IP family.
```
wpross, line 65: mh, would be nice to have a unit test for this fctn, but it seems there are no tests in this folder at all 🤔

pkg/controller/infrastructure/infraflow/reconcile.go
```diff
@@ -306,7 +306,7 @@ func (fctx *FlowContext) ensureSubnet(ctx context.Context) error {
 		Name:           fctx.defaultSubnetName(),
 		NetworkID:      networkID,
 		IPVersion:      4,
-		DNSNameservers: fctx.cloudProfileConfig.DNSServers,
+		DNSNameservers: filterDNSServersByIPFamily(fctx.cloudProfileConfig.DNSServers, gardencorev1beta1.IPFamilyIPv4),
```
wpross, line 309: I'm wondering if it would be good to log a warning message or similar when there's no matching DNSNameserver? E.g.
```
filtered := filterDNSServersByIPFamily(fctx.cloudProfileConfig.DNSServers, gardencorev1beta1.IPFamilyIPv6)
  if len(fctx.cloudProfileConfig.DNSServers) > 0 && len(filtered) == 0 {
      log.Info("WARNING: cloud profile DNS servers contain no IPv6 addresses; IPv6 subnets will have no DNS nameservers configured")
  }
```

pkg/controller/infrastructure/infraflow/utils.go
hebelsan: I've created a new utils_test.go file and added tests for this function.

pkg/controller/infrastructure/infraflow/reconcile.go
hebelsan: Yes, I think this would be useful, but logging this information on every reconcile would also create quite a bit of noise.
I think the best approach would be to add a warning to the Shoot or infrastructure resource so that it becomes visible in the dashboard, although that would be a larger piece of work.
Alternatively, we could log it at the debug level.
@kon-angelo wdyt?

kon-angelo: But aren't dns servers in the cloudprofile optional in the first place ? You would have a shoot status warning for something that is completely within operational parameters. 

For whom do you think this is crucial information ? Is this useful info for the operators who they themselves crafted the cloudprofile ? 

You can also consider adding that to the subnet info in the provider status.

fzakfeld: I found that workers fail to join if DNS is broken / missing. It of course does not need to be dual stack DNS. Lots of users will keep using IPv4-only shoots, so making IPv6 nameservers mandatory also doesn't seem right

wpross: @kon-angelo you're right, they are completely optional, so then a warning makes no sense.
@fzakfeld the issue was that there were dns servers configured, but with the wrong ip version. This fix should solve it by not returning any dns servers for a specific ip version if there are none.

wpross (approved):
/lgtm

PR #1349, opened 2026-05-27
[port] Enable caching in build pipelines

matthias-horne:
Port of gardener/gardener-extension-provider-azure#1530 from azure.

## Changes

- Reorder build steps: move `trusted-checkout` before `setup-go` so `setup-go` can use the checked-out `go.mod`/`go.sum` for cache key generation
- Upgrade `setup-go` from v5 to v6

/area dev-productivity
/kind enhancement

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -121,12 +121,12 @@ jobs:
     permissions:
       contents: read
     steps:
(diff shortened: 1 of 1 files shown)


wpross (approved):
/lgtm

PR #1351, opened 2026-05-28
`Fix:` add share client worker cidr fall back

hebelsan:

/area control-plane
/kind bug
/platform openstack

https://github.com/gardener/gardener-extension-provider-openstack/pull/1326 introduced a race condition for existing shoots.
The change added a new subnetCIDR status field that is consumed during control plane reconciliation. 
However, the infrastructure reconciliation may not have populated the infrastructure status yet when the control plane reconciliation reads it, leading to a race condition for existing shoots.
For new shoots, the infra controller always writes CIDR into the subnet status from the very first reconcile so nodesSubnet.CIDR will always be set before the control plane can meaningfully reconcile.

Fixes #

**Release note**:

```bugfix operator
Fix CSI manila share client value race condition in control plane reconciliation
```

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -923,6 +923,10 @@ func (vp *valuesProvider) addCSIManilaValues(
 		"clusterID": cp.Namespace,
 	}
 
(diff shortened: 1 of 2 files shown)


hebelsan:
Reproduced the issue and verified the fix in my local setup ✅

kon-angelo (approved):
I see. We missed the fact that existing shoots may reconcile CP before reconciling the infrastructure in between using the new field.

PR #1352, opened 2026-05-28
Fix: prevent nil pointer panic in machine image lookup

Roncossek:

/area robustness
/kind bug
/platform openstack

Fixes a nil pointer dereference panic in `FindImageInWorkerStatus` that crashed the worker controller during reconcile and delete of half-initialized workers (e.g. workers that never finished their first reconciliation, so `providerStatus.machineImages` is empty, while the referenced image is also no longer present in the CloudProfile).

- Change `FindImageInWorkerStatus` `architecture` parameter from `*string` to `string`; default empty to `amd64`.
- Pass an architecture explicitly from `selectMachineImageForWorkerPool` (previously hard-coded `nil`, which both caused the panic and prevented any legacy-format match from succeeding).
- Update tests accordingly.

Fixes #

cc @hebelsan 

**Release note**:
```bugfix operator
Fix a nil pointer dereference panic in the worker controller that could prevent reconciliation or deletion of workers whose `providerStatus.machineImages` was empty and whose referenced machine image was missing from the CloudProfile.
```


hebelsan (approved):
/lgtm

PR #1359, opened 2026-06-01
Fix: credentials validation to support InternalSecret type

hebelsan:

/area quality
/kind bug
/platform openstack

This PR adds the missing credentials validation to support InternalSecret type

Fixes #

**Release note**:

```bugfix operator
Add credentials validation to support InternalSecret type
```

### DIFF
--- a/pkg/admission/validator/credentialsbinding.go
+++ b/pkg/admission/validator/credentialsbinding.go
@@ -9,7 +9,9 @@ import (
 	"fmt"
 
 	extensionswebhook "github.com/gardener/gardener/extensions/pkg/webhook"
+	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	"github.com/gardener/gardener/pkg/apis/security"
+	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -48,15 +50,16 @@ func (cb *credentialsBinding) Validate(ctx context.Context, newObj, oldObj clien
 
 	// Explicitly use the client.Reader to prevent controller-runtime to start Informer for Secrets
 	// under the hood. The latter increases the memory usage of the component.
-	var credentialsKey = client.ObjectKey{Namespace: credentialsBinding.CredentialsRef.Namespace, Name: credentialsBinding.CredentialsRef.Name}
-	switch {
-	case credentialsBinding.CredentialsRef.APIVersion == corev1.SchemeGroupVersion.String() && credentialsBinding.CredentialsRef.Kind == "Secret":
-		secret := &corev1.Secret{}
-		if err := cb.apiReader.Get(ctx, credentialsKey, secret); err != nil {
-			return err
-		}
+	credentials, err := kutil.GetCredentialsByObjectReference(ctx, cb.apiReader, credentialsBinding.CredentialsRef)
+	if err != nil {
+		return err
+	}
 
-		return openstackvalidation.ValidateCloudProviderSecret(secret, field.NewPath("secret"), false).ToAggregate()
+	switch creds := credentials.(type) {
+	case *corev1.Secret:
+		return openstackvalidation.ValidateCloudProviderSecret(creds, field.NewPath("secret"), false).ToAggregate()
+	case *gardencorev1beta1.InternalSecret:
+		return openstackvalidation.ValidateCloudProviderSecretData(creds.Data, field.NewPath("secret"), false).ToAggregate()
 	default:
 		return fmt.Errorf("unsupported credentials reference: version %q, kind %q", credentialsBinding.CredentialsRef.APIVersion, credentialsBinding.CredentialsRef.Kind)
 	}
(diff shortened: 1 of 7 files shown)


pkg/admission/validator/shoot.go
```diff
@@ -247,17 +252,27 @@ func (s *shoot) getCloudProviderSecretForShoot(ctx context.Context, shoot *core.
...
+	if err != nil {
 		return nil, err
 	}
 
-	return secret, nil
+	switch creds := credentials.(type) {
+	case *corev1.Secret:
+		return creds, nil
+	case *gardencorev1beta1.InternalSecret:
+		return &corev1.Secret{
```
wpross: I'm wondering if returning a secret for the `InternalSecret` could be omitted by doing it in a similar way as in case of the `openstackvalidation.ValidateCloudProviderSecretData()`? So basically creating a new fctn `ExtractCredentialsFromData` additionally to the https://github.com/gardener/gardener-extension-provider-openstack/blob/master/pkg/openstack/credentials.go#L48 since the `data` field is the only one needed in `https://github.com/gardener/gardener-extension-provider-openstack/blob/master/pkg/admission/validator/shoot.go#L86`?
Maybe cleaner then wrapping the `InternalSecret` into a `Secret`?

pkg/admission/validator/credentialsbinding.go
```diff
@@ -48,15 +50,16 @@ func (cb *credentialsBinding) Validate(ctx context.Context, newObj, oldObj clien
...
-		}
+	credentials, err := kutil.GetCredentialsByObjectReference(ctx, cb.apiReader, credentialsBinding.CredentialsRef)
+	if err != nil {
+		return err
+	}
 
-		return openstackvalidation.ValidateCloudProviderSecret(secret, field.NewPath("secret"), false).ToAggregate()
+	switch creds := credentials.(type) {
+	case *corev1.Secret:
+		return openstackvalidation.ValidateCloudProviderSecret(creds, field.NewPath("secret"), false).ToAggregate()
```
wpross: I would make it consistent and in both cases use `ValidateCloudProviderSecretData`.

pkg/admission/validator/credentialsbinding_test.go
```diff
@@ -122,5 +122,49 @@ var _ = Describe("CredentialsBinding validator", func() {
...
+			It("should return err if it fails to get the corresponding InternalSecret", func() {
+				apiReader.EXPECT().Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, gomock.AssignableToTypeOf(&gardencorev1beta1.InternalSecret{})).Return(fakeErr)
+
+				err := credentialsBindingValidator.Validate(ctx, credentialsBinding, nil)
+				Expect(err).To(MatchError(fakeErr))
+			})
+
+			It("should return err when the corresponding InternalSecret is not valid", func() {
+				apiReader.EXPECT().Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, gomock.AssignableToTypeOf(&gardencorev1beta1.InternalSecret{})).
+					DoAndReturn(func(_ context.Context, _ client.ObjectKey, obj *gardencorev1beta1.InternalSecret, _ ...client.GetOption) error {
```
wpross, line 145: I personally prefer the `SetArg()` for readability, but it's consistent to use the `DoAndReturn` here 🤷‍♂️

hebelsan: For now I would leave it as is.

hebelsan:
Successfully tested shoot creation in my local setup with an internal secret in the credential binding.

pkg/openstack/credentials.go
```diff
@@ -60,28 +67,27 @@ func ExtractCredentials(secret *corev1.Secret, allowDNSKeys bool) (*Credentials,
...
-	password := getOptional(secret, Password, altPasswordKey)
-	applicationCredentialID := getOptional(secret, ApplicationCredentialID, altApplicationCredentialID)
-	applicationCredentialName := getOptional(secret, ApplicationCredentialName, altApplicationCredentialName)
-	applicationCredentialSecret := getOptional(secret, ApplicationCredentialSecret, altApplicationCredentialSecret)
-	authURL := getOptional(secret, AuthURL, altAuthURLKey)
-	caCert := getOptional(secret, CACert, altCABundleKey)
-
-	err = ValidateSecrets(userName, password, applicationCredentialID, applicationCredentialName, applicationCredentialSecret)
-	if err != nil {
-		return nil, fmt.Errorf("%w in secret %s/%s", err, secret.Namespace, secret.Name)
```
wpross, line 84: I guess that passing `secret.Namespace, secret.Name` through in the error message helps identifying which secret caused the error.

pkg/openstack/credentials.go
```diff
@@ -120,36 +126,36 @@ func ValidateSecrets(userName, password, appID, appName, appSecret string) error
...
 
-// getRequired checks if the provided map has a valid value for a corresponding key.
-func getRequired(secret *corev1.Secret, key string, altKey *string) (string, error) {
-	value, ok := secret.Data[key]
+// getRequiredFromData checks if the provided map has a valid value for a corresponding key.
+func getRequiredFromData(data map[string][]byte, key string, altKey *string) (string, error) {
+	value, ok := data[key]
 	if !ok {
 		if altKey == nil {
-			return "", fmt.Errorf("missing %q data key in secret %s/%s", key, secret.Namespace, secret.Name)
```
wpross, line 141: I guess that passing `secret.Namespace, secret.Name` through in this error message and the others in `getRequired()` helps identifying which secret caused the error.

wpross (changes requested):

pkg/openstack/credentials.go
hebelsan: this one is called inside of `ExtractCredentials` from above and therefore will contain name and namespace.

wpross (approved):
/lgtm

PR #1364, opened 2026-06-08
Add tagging documentation

hebelsan:

/area documentation
/kind enhancement
/platform openstack

This PR adds a new documentation file that explains which resources get which tags.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/docs/usage/openstack-resource-tagging.md
+++ b/docs/usage/openstack-resource-tagging.md
@@ -1,12 +1,12 @@
 # OpenStack Resource Tagging
 
-This document describes which OpenStack resources are tagged by
-`gardener-extension-provider-openstack` and what tags are applied to each resource type.
-At the moment, tagging is implemented only for worker node virtual machines.
+This document describes which OpenStack resources are annotated by
+`gardener-extension-provider-openstack` and what metadata is applied to each resource type.
+At the moment, metadata tagging is implemented only for worker node virtual machines.
 
 ## Overview
 
-The extension applies metadata tags to OpenStack resources for two primary purposes:
+The extension sets server metadata on OpenStack resources for two primary purposes:
 
 1. **Ownership identification** — marking resources as Gardener-managed so they
    can be found, filtered, and reconciled correctly.
@@ -18,10 +18,10 @@ The extension applies metadata tags to OpenStack resources for two primary purpo
 
 ### Virtual Machines (Worker Nodes)
 
-Worker node VMs receive metadata tags derived from the worker pool configuration
+Worker node VMs receive server metadata entries derived from the worker pool configuration
 in the Shoot spec.
 
-| Tag Key | Value | Source |
+| Metadata Key | Value | Source |
 |---|---|---|
 | `kubernetes.io-cluster-{technicalID}` | `"1"` | Shoot technical ID |
 | `kubernetes.io-role-node` | `"1"` | Static |
@@ -46,7 +46,7 @@ spec:
           value: my-value
 ```
 
-The resulting VM metadata tags are:
+The resulting VM server metadata entries are:
 
 ```text
 kubernetes.io-cluster-shoot--my-project--my-cluster: "1"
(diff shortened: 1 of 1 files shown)


docs/usage/openstack-resource-tagging.md
```diff
@@ -0,0 +1,100 @@
+# OpenStack Resource Tagging
```
wpross, line 1: This sounds very broad, like all resources. Would be worth mentioning that so far, only VMs are tagged and other resources are TODOs.

docs/usage/openstack-resource-tagging.md
```diff
@@ -0,0 +1,100 @@
...
+    triggerRollingOnUpdate: true
+```
+
+## Tag Sanitization
+
+OpenStack server metadata keys do not allow certain characters that are common
+in Kubernetes label keys. Worker pool label keys and machine label keys are
+sanitized before being applied as OpenStack metadata tags.
+
+**Allowed characters in tag keys:** `a-zA-Z0-9 -_:.` (alphanumeric, hyphen,
```
wpross: A link to the mentioned constraint in openstack docs would be nice.
I could only find https://docs.openstack.org/api-ref/compute/
stating a max 255 chars for key+value though...

docs/usage/openstack-resource-tagging.md
```diff
@@ -0,0 +1,100 @@
...
+underscore, colon, period, space)
+
+Any character outside this set is replaced with a hyphen (`-`).
+
+**Examples:**
+
+| Original Kubernetes Label Key | Sanitized OpenStack Tag Key |
```
wpross: what if keys end up being the same after sanitization? Which one wins?
And what if they are defined within different sections (e.g. machineLabels vs. workers.labels)?

wpross (changes requested):

docs/usage/openstack-resource-tagging.md
hebelsan: very good question!
Added some lines for clarification.

docs/usage/openstack-resource-tagging.md
```diff
@@ -83,7 +84,7 @@ in Kubernetes label keys. Worker pool label keys and machine label keys are
 sanitized before being applied as OpenStack metadata tags.
 
 **Allowed characters in tag keys:** `a-zA-Z0-9 -_:.` (alphanumeric, hyphen,
-underscore, colon, period, space)
+underscore, colon, period, space), see [OpenStack documentation](https://docs.openstack.org/api-ref/compute/#server-tags-servers-tags).
```
wpross: mh, from the linked doc it seems only `/` and `,` is not allowed, everything else is

also I'm wondering if https://docs.openstack.org/api-ref/compute/#create-or-update-metadata-items is the correct link, so basically if tags are set or metadata?
see e.g. https://github.com/gardener/machine-controller-manager-provider-openstack/blob/master/pkg/driver/executor/executor.go#L712

docs/usage/openstack-resource-tagging.md
```diff
@@ -98,3 +99,26 @@ Any character outside this set is replaced with a hyphen (`-`).
 
 Note that only tag **keys** are sanitized; tag **values** are passed through
 unchanged.
+
+### Collision Handling
+
+If two different original keys sanitize to the same tag key, the last value
+written wins.
+if both are present in the same label source, one will silently overwrite the
```
wpross: NIT: "If" with capital "I"

wpross (changes requested):

docs/usage/openstack-resource-tagging.md
hebelsan: You're right, it's actually the server metadata that we set. 
I've updated the wording and the link.
However, it appears that we're stricter than OpenStack, which allows almost any value.

wpross: Do you know why we're so strict here?

wpross (approved):
/lgtm

PR #1366, opened 2026-06-10
Use the correct API version for MutatingAdmissionPolicy

DockToFuture:
`MutatingAdmissionPolicy` graduates from `v1alpha1` → `v1beta1` in Kubernetes 1.34 and to `v1` (GA) in 1.36. Render the chart with the correct API version based on the shoot's Kubernetes version, and align the enable/disable logic accordingly:

- `< 1.34`: `v1alpha1` only — `v1beta1` does not exist on these clusters, so a `v1beta1` RuntimeConfig opt-in is ignored
- `>= 1.34 and < 1.36`: `v1beta1`, beta but disabled by default — requires explicit feature gate + RuntimeConfig opt-in
- `>= 1.36`: `v1`, always enabled (feature gate locked on)

Also fixes a pre-existing object-name mismatch between the chart template (`block-calico-network-unavailable`) and the registered chart objects (formerly `calico-node-status-patch`).

### DIFF
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -12,6 +12,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
 	"github.com/gardener/gardener/extensions/pkg/controller/controlplane/genericactuator"
 	extensionssecretmanager "github.com/gardener/gardener/extensions/pkg/util/secret/manager"
@@ -24,8 +25,10 @@ import (
 	kutil "github.com/gardener/gardener/pkg/utils/kubernetes"
 	secretutils "github.com/gardener/gardener/pkg/utils/secrets"
 	secretsmanager "github.com/gardener/gardener/pkg/utils/secrets/manager"
+	versionutils "github.com/gardener/gardener/pkg/utils/version"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	admissionregistrationv1alpha1 "k8s.io/api/admissionregistration/v1alpha1"
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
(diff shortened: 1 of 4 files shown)


pkg/controller/controlplane/valuesprovider.go
```diff
@@ -871,26 +883,69 @@ func (vp *valuesProvider) isUsingCalico(cluster *extensionscontroller.Cluster) b
 	return *cluster.Shoot.Spec.Networking.Type == "calico"
 }
 
+// constraintK8sGreaterEqual136 is a version constraint for Kubernetes versions >= 1.36.
+var constraintK8sGreaterEqual136 = versionutils.MustNewConstraint(">= 1.36-0")
```
wpross, line 888: I would add a TODO cleanup comment to use the `versionutils` version from g/g (like in case of the 134 check) once it is available to not forget it.

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -871,26 +883,69 @@ func (vp *valuesProvider) isUsingCalico(cluster *extensionscontroller.Cluster) b
...
+	if constraintK8sGreaterEqual136.Check(k8sVersion) {
+		return "v1"
+	}
+
+	// For K8s >= 1.34, MutatingAdmissionPolicy is beta
+	if versionutils.ConstraintK8sGreaterEqual134.Check(k8sVersion) {
+		return "v1beta1"
+	}
+
+	// For older versions, prefer v1beta1 over v1alpha1 when explicitly enabled
```
wpross: does this work to use `v1beta1` for k8s versions < `v1.34`? From the PR's description, it is not.

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -871,26 +883,69 @@ func (vp *valuesProvider) isUsingCalico(cluster *extensionscontroller.Cluster) b
...
+
+	rc := cluster.Shoot.Spec.Kubernetes.KubeAPIServer.RuntimeConfig
+
+	// For K8s >= 1.34 and < 1.36, MutatingAdmissionPolicy is in beta but disabled by default.
+	// Users must explicitly enable the feature gate and opt in via RuntimeConfig.
+	if versionutils.ConstraintK8sGreaterEqual134.Check(k8sVersion) {
+		return rc["admissionregistration.k8s.io/v1beta1"]
+	}
+
+	return rc["admissionregistration.k8s.io/v1alpha1"] || rc["admissionregistration.k8s.io/v1beta1"]
```
wpross: see https://github.com/gardener/gardener-extension-provider-openstack/pull/1366/changes#r3414440520

wpross (changes requested):

DockToFuture:
Hello @wpross, I've update the PR. Will do the same for aws and gcp to introduce the same behaviour.

wpross (approved):
/lgtm

PR #1401, opened 2026-07-31
[GEP-36] `SelfHostedShootExposure`

tobschli:

/area quality
/kind enhancement
/platform openstack

This PR implements the `SelfHostedShootExposure` controller according to its GEP, but creating and managing a LB directly on openstack.

In addition to that, I also included smaller fixes that were necessary to make the extension work in the self hosted shoot scenario.

Part of https://github.com/gardener/gardener/issues/13602

**Release note**:

```feature operator
`SelfHostedShootExposure` is now supported by this extension.
```

## Summary by CodeRabbit

- **New Features**
  - Added OpenStack support for self-hosted shoot exposure, including load balancer provisioning, pool synchronization, floating IP assignment, and control-plane access rules.
  - Added configurable exposure reconciliation concurrency, defaulting to five concurrent operations.
- **Bug Fixes**
  - Improved recovery and cleanup of failed or partially deleted exposure resources.
  - Corrected CSI node cloud-provider configuration handling.
  - Improved server-group naming and cleanup when shoot identity information is available.
- **Tests**
  - Added comprehensive coverage for exposure provisioning, updates, deletion, recovery, and error handling.


kon-angelo:
@tobschli (cc:@ScheererJ) - Can you please state why you decided to run a LB reconciler from the extension and not used a CCM ? Even a reduced responsibility "service-only". I just want to understand because I assume this work will be replicated with other providers ? 

Was it maybe an openstack only implementations ? Would you evaluate CCMs if feasible in other providers instead of the self-exposure?

ScheererJ:
> @tobschli (cc:@ScheererJ) - Can you please state why you decided to run a LB reconciler from the extension and not used a CCM ? Even a reduced responsibility "service-only". I just want to understand because I assume this work will be replicated with other providers ?
> 
> Was it maybe an openstack only implementations ? Would you evaluate CCMs if feasible in other providers instead of the self-exposure?

This is described in the [alternatives section of the enhancement proposal](https://github.com/gardener/enhancements/tree/main/geps/0036-self-hosted-shoot-exposure#alternatives). Does this make sense from your point of view?

Yes, we want to implement this also for other infrastructure providers.

As mentioned in the [enhancement proposal](https://github.com/gardener/enhancements/tree/main/geps/0036-self-hosted-shoot-exposure#examples-of-possible-extension-implementations), in an L2 network like provided by openstack other implementations are possible. The load balancer approach is the more general one. So other provider implementations will likely be similar to this PR. However, on openstack we could use a different approach.

pkg/controller/selfhostedshootexposure/helpers_test.go
hebelsan, line 1: Shall we remove this file since it contains no tests?

pkg/controller/selfhostedshootexposure/helpers.go
```diff
@@ -0,0 +1,206 @@
...
+// control-plane machines on the exposure port. The amphora backend address is not discoverable, so
+// the rule cannot be scoped to a single source; it opens the exposure port (the TLS-protected
+// kube-apiserver) from anywhere.
+func securityGroupRuleOpts(exposure *extensionsv1alpha1.SelfHostedShootExposure, securityGroupID string, family gardencorev1beta1.IPFamily) rules.CreateOpts {
+	etherType := rules.EtherType4
+	prefix := "0.0.0.0/0"
+	if family == gardencorev1beta1.IPFamilyIPv6 {
+		etherType = rules.EtherType6
+		prefix = "::/0"
+	}
```
hebelsan, line 74: The rule opens the exposure port to 0.0.0.0/0 on the nodes security group, justified by:
> The amphora backend port address is not discoverable via the load balancer API, so the rule cannot be scoped to a single source; it opens the exposure port (the TLS-protected  kube-apiserver) from anywhere.

My assumption is:
HAProxy on the amphora terminates the client connection and re-originates a new one to the member, so the source IP the member sees is always the amphora's, never the client's. The members never see arbitrary source IPs on this port.
So the rule doesn't need to admit "anyone who might use the apiserver" — only the amphora. It's correct that the amphora's exact IP isn't discoverable, but its member-facing port lives on the VIP subnet, and this controller sets `VipSubnetID == nodesSubnet.ID`. That means the amphora's source IP is guaranteed to be inside the nodes-subnet CIDR, which we already have from `infraStatus.Networks.Subnets`.

So my question is:
If amphoras are always on the VIP (= nodes) subnet, could we scope `RemoteIPPrefix` to the nodes-subnet CIDR instead?

ScheererJ: Your reasoning sound valid. I share your assumption that a range would be sufficient and it does not have to be the exact IP address. This would certainly reduce the potential attack vector.

pkg/controller/selfhostedshootexposure/helpers_test.go
ScheererJ: Agreed. Let's remove the file.

pkg/controller/selfhostedshootexposure/helpers.go
ScheererJ: @hebelsan PTAL.

hebelsan:
@ScheererJ thanks for addressing the comments! 🚀
Could you please check `make verify`, it currently fails with:
```
pkg/controller/selfhostedshootexposure/actuator.go:172:58: printf: fmt.Errorf format %w has arg purpose of wrong type github.com/gardener/gardener-extension-provider-openstack/pkg/apis/openstack.Purpose
```
Also, please check code rabbits comments.

rfranzke:
Can we merge this?

PR #1406, opened 2026-08-06
Reduce unnecessary workflow runs

matthias-horne:

/area cost
/kind cleanup
/platform openstack

Several small improvements to the GitHub actions:
- removed several unnecessary trigger events.
- release milestone check pulls version file through API instead of checking out the entire repository.
- add concurrency group for non-release workflow. This will cancel workflows in the `pull_request` context, when the workflow is triggered in the `pull_request_target` context. Since both are triggered in quick succession for trusted maintainers did should almost half our workflow run time.
- provide a nice name for PR trust helper workflow.

Fixes #

**Release note**:

```other operator

```

## Summary by CodeRabbit

- **Chores**
  - Improved pull request automation by reducing unnecessary workflow runs when pull requests are edited.
  - Added safeguards to cancel outdated checks when newer changes are submitted.
  - Streamlined release milestone processing to use the latest version information and avoid redundant triggers.
  - Updated automation behavior for newly opened, reopened, or synchronized pull requests.

### DIFF
--- a/.github/workflows/pullrequest-trust-helper.yaml
+++ b/.github/workflows/pullrequest-trust-helper.yaml
@@ -1,8 +1,8 @@
+name: Pull Request Trust Helper
 on:
   pull_request_target:
     types:
     - opened
-    - edited
     - reopened
     - synchronize
 
--- a/.github/workflows/release-milestone.yaml
+++ b/.github/workflows/release-milestone.yaml
@@ -4,9 +4,6 @@ on:
     types:
     - opened
     - reopened
-    - synchronize
-    - labeled
-    - unlabeled
     - milestoned
     - demilestoned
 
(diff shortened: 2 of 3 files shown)


.github/workflows/pullrequest-trust-helper.yaml
```diff
@@ -1,8 +1,8 @@
+name: Pull Request Trust Helper
 on:
   pull_request_target:
     types:
     - opened
-    - edited
```
hebelsan, line 5: Can we confirm this is okay?

The doc state the following:
https://github.com/gardener/cc-utils/blob/e6d454eda2d8b8e5894c5498d6136b5c5b81d43e/.github/workflows/pullrequest-trust-helper.yaml#L18

And let's think of the following example:
  1. Open a PR targeting a non-protected branch → trust helper runs, you get ok-to-test automatically (since it's a low-trust target)
  2. Edit the PR to retarget to master → without edited, the trust helper doesn't re-run, the ok-to-test label stays
  3. Now CI runs on master-targeting code with a label that was granted under different circumstances

.github/workflows/release-milestone.yaml
```diff
@@ -4,9 +4,6 @@ on:
     types:
     - opened
     - reopened
-    - synchronize
-    - labeled
-    - unlabeled
```
hebelsan, line 9: I understand why we'd remove those, but are we sure this is okay?

Consider the case where I create a new milestone and want all existing PRs to be blocked from merging until they're assigned the correct milestone.

Previously, labeled, unlabeled, and synchronized implicitly covered this scenario. Events like adding `ok-to-test`, `lgtm`, or `approved` labels happen frequently before a merge, so the check would reliably re-run before anything could be merge.

With this change, wouldn't existing PRs be able to merge without ever re-running the check?

Issue #947, opened 2025-01-09
Enable Seperate CIDR for LB and Worker nodes definition in gardener


hariprasath91:
**How to categorize this issue?**

/area networking
/kind enhancement
/platform openstack

**We want to enable below enhancement where ,**
1. Need to enable a seperate CIDR for LB for the shoot cluster.
2. Need to enable a seperate CIDR for nodes on the shoot 
3. Enable possibility of enabling internal loadbalancer.
4. Enable the possibility of enabling external loadbalancer.

**Why is this needed**
We need this feature to isolate the access to the customer projects when the BGP VPN is enabled. We dont want to enable the access to entire cluster CIDR rather only to LB CIDR alone where the connections can be enabled,

PR #1417, opened 2026-08-25
Internal load balancer

stblatzheim:

/area networking
/area control-plane
/kind enhancement
/kind api-change
/platform openstack

This PR adds the optional `cloudControllerManager.internalLoadBalancer` field to the OpenStack `ControlPlaneConfig`.

The value is propagated to the OpenStack cloud-controller-manager configuration as:

```ini
[LoadBalancer]
internal-lb=true
```

This allows users to configure on a per-Shoot basis whether all Kubernetes Services of type `LoadBalancer` should be provisioned as internal load balancers without floating IPs.

Example:

```yaml
spec:
  provider:
    controlPlaneConfig:
      kind: ControlPlaneConfig
      loadBalancerProvider: ovn
      cloudControllerManager:
        internalLoadBalancer: true
```

The option defaults to `false`, preserving the existing behavior for all current and newly created Shoots that do not explicitly enable it.

Partially addresses #947.

This change exposes the existing `internal-lb` configuration option of the OpenStack cloud-controller-manager. It does not introduce custom load balancer behavior in the Gardener extension.

When enabled, the setting applies to all Services of type `LoadBalancer` in the respective Shoot, not only to a particular ingress controller.

The setting can also be changed for existing Shoots. Enabling it may cause the OpenStack cloud-controller-manager to detach floating IPs from existing load balancers during reconciliation.

**Release note**:

```feature user
OpenStack Shoot clusters can now configure `cloudControllerManager.internalLoadBalancer` in their `ControlPlaneConfig`. When enabled, all Services of type `LoadBalancer` are provisioned as internal load balancers without floating IPs.
```

## Summary by CodeRabbit

* **New Features**
  * Added an option to configure all `LoadBalancer` Services as internal.
  * Prevented external load balancer creation when internal load balancing is enabled.
  * Added configuration support across control plane settings and deployment values.

* **Documentation**
  * Documented the optional setting, its default behavior, and its impact on existing load balancers.

### DIFF
--- a/pkg/apis/openstack/v1alpha1/types_controlplane.go
+++ b/pkg/apis/openstack/v1alpha1/types_controlplane.go
@@ -38,9 +38,9 @@ type CloudControllerManagerConfig struct {
 	// +optional
 	FeatureGates map[string]bool `json:"featureGates,omitempty"`
 	// InternalLoadBalancer configures all Services of type LoadBalancer
-    // to be internal and prevents creating external load balancers.
-    // +optional
-    InternalLoadBalancer bool `json:"internalLoadBalancer,omitempty"`
+	// to be internal and prevents creating external load balancers.
+	// +optional
+	InternalLoadBalancer bool `json:"internalLoadBalancer,omitempty"`
 }
 
 // Storage contains configuration for storage in the cluster.
--- a/pkg/apis/openstack/v1alpha1/types_controlplane.go
+++ b/pkg/apis/openstack/v1alpha1/types_controlplane.go
@@ -37,6 +37,10 @@ type CloudControllerManagerConfig struct {
 	// FeatureGates contains information about enabled feature gates.
 	// +optional
 	FeatureGates map[string]bool `json:"featureGates,omitempty"`
+	// InternalLoadBalancer configures all Services of type LoadBalancer
+    // to be internal and prevents creating external load balancers.
+    // +optional
+    InternalLoadBalancer bool `json:"internalLoadBalancer,omitempty"`
 }
 
 // Storage contains configuration for storage in the cluster.
--- a/pkg/controller/controlplane/valuesprovider.go
+++ b/pkg/controller/controlplane/valuesprovider.go
@@ -547,8 +547,7 @@ func getConfigChartValues(
 	}
 
 	if cpConfig.CloudControllerManager != nil {
-		values["internalLoadBalancer"] =
-			cpConfig.CloudControllerManager.InternalLoadBalancer
+		values["internalLoadBalancer"] = cpConfig.CloudControllerManager.InternalLoadBalancer
 	}
 
 	if !isUsingOverlay {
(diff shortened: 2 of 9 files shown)


pkg/apis/openstack/v1alpha1/types_controlplane.go
```diff
@@ -37,6 +37,10 @@ type CloudControllerManagerConfig struct {
 	// FeatureGates contains information about enabled feature gates.
 	// +optional
 	FeatureGates map[string]bool `json:"featureGates,omitempty"`
+	// InternalLoadBalancer configures all Services of type LoadBalancer
+    // to be internal and prevents creating external load balancers.
+    // +optional
+    InternalLoadBalancer bool `json:"internalLoadBalancer,omitempty"`
```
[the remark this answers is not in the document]
stblatzheim: added needed changes

hebelsan:
@stblatzheim thanks for the contribution! 🚀
Please run `make generate verify` to format, auto generate API + mock code and run the unit tests.

_pkg/controller/controlplane/valuesprovider_test.go_ asserts the config chart values via an exact Equal(configChartValues) match. Please add a test exercising the new field.

_docs/usage/usage.md_ documents cloudControllerManager.featureGates but not the new internalLoadBalancer field. Given the PR is tagged `/kind api-change` and the description warns about the non-trivial side effect (enabling it on existing Shoots can detach floating IPs during reconciliation), that caveat belongs in the user docs, not just the PR body.

stblatzheim:
Hey @hebelsan,
thanks for the review. I ran make generate verify and pushed the resulting generated and formatted changes. 
I also added a dedicated values-provider test with internalLoadBalancer enabled and documented the option, including the floating-IP detachment caveat for existing Shoots, in docs/usage/usage.md.

hebelsan:
@stblatzheim seems like `make verify` still fails.

stblatzheim:
Hey @hebelsan,

fixed and checke again:

Summary:
  Gosec  : dev
  Files  : 122
  Lines  : 16507
  Nosec  : 5
  Issues : 0

pkg/controller/controlplane/valuesprovider.go
```diff
@@ -546,6 +546,11 @@ func getConfigChartValues(
 		"internalNetworkName": infraStatus.Networks.Name,
 	}
 
+	if cpConfig.CloudControllerManager != nil {
+		values["internalLoadBalancer"] =
+			cpConfig.CloudControllerManager.InternalLoadBalancer
```
hebelsan: Small nit: This wraps a short assignment across two lines oddly but it fits on one line comfortably.

hebelsan (commented):
Thanks for addressing my previous comments! 😊
The PR looks good now, just one small nit.

pkg/controller/controlplane/valuesprovider.go
stblatzheim: Thanks for the hint :)

Solved with the latest commit.

PR #1419, opened 2026-08-27
Fix: don't read bastion floating IP from Nova server addresses

hebelsan:

/area robustness
/kind enhancement
/platform openstack

The bastion controller failed to become ready with: `Error reconciling Bastion: no IP found: no public IP found`
 **Root cause:**
 The reconcile flow associates the floating IP to the bastion at the **Neutron port level** via `UpdateFIPWithPort`. Afterwards, `GetIPs` tried to rediscover that public IP by parsing Nova's `server.Addresses` map for an entry of type `OS-EXT-IPS:type == "floating"`.
When a floating IP is attached directly through Neutron, Nova's server `Addresses` view does not reliably expose the floating entry. As a result `GetIPs` returned `no public IP found`, and the bastion never reached a ready state regardless of requeues. This is not a floating-IP quota issue.
**Fix:**
The floating IP object is already known from `ensurePublicIPAddress`, so there is no need to read it back out of Nova

Fixes #

**Release note**:

```bugfix operator
Don't read bastion floating IP from Nova server addresses
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Bastion endpoints now use the correct floating IP for public access.
  * Private endpoint discovery now reliably returns the server’s fixed private IP.
  * Improved handling when public IP information is unavailable.

* **Tests**
  * Updated bastion creation checks to validate private and floating IPs separately.

### DIFF
--- a/pkg/controller/bastion/actuator.go
+++ b/pkg/controller/bastion/actuator.go
@@ -66,9 +66,12 @@ func deleteBastionInstance(ctx context.Context, client openstackclient.Compute, 
 	return client.DeleteServer(ctx, id)
 }
 
-// GetIPs returns the first found private and public IPs for the given server and options.
-func GetIPs(s servers.Server, opts Options) (string, string, error) {
-	var privateIP, publicIP string
+// GetPrivateIP returns the first found private (fixed) IP for the given server.
+// The public (floating) IP is not read from the server addresses because the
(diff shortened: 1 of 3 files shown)


matthias-horne (commented):
/lgtm
/approve

PR #1428, opened 2026-09-03
Add per-region settings for storageClasses

### LINKED ISSUE #1427: Per-region storageClasses parameters

/area storage
/kind enhancement
/platform openstack

I propose adding an optional `regions` field for the `StorageClassDefinition` struct that allows two things:
 * overwrite `parameters.type` to map to a different openstack volume type for a particular region
 * make a `StorageClass` unavailable in a region (because no related openstack volume type exists)

Example:
```
storageClasses:
  - name: example-sc
    default: false
    provisioner: cinder.csi.openstack.org
    volumeBindingMode: WaitForFirstConsumer
    parameters:
      type: storage_premium_perf0   # used in all other regions
    regions:
    - name: europe-west
      parameters:
        type: premium            # used only in this one region
    - name: asia
      unavailable: true          # this storageClass does not get created at all in this region
```
**Why is this needed**:
We have different volume type names across openstack regions, and need to express that difference in the cloudprofile to create storageClasses that are actually working. It's not in our hands to keep them the same name in openstack itself, and some regions genuinely lack some storage tiers. This is currently not possible with providerConfig.storageClasses.

lotharbach:

/area storage
/kind enhancement
/platform openstack

The openstack `volume-type` behind a `storageClass` can differ per region, and some
regions do not offer a matching `volume-type` at all. Neither case could be expressed in
`providerConfig.storageClasses` so far.

This adds an optional `regions` list to `StorageClassDefinition`. Each entry selects a
region by name and either overrides the storageClass `parameters` key by key, or marks the
storageClass as unavailable so that it is not created for shoots in that region:

```yaml
storageClasses:
- name: example-sc
  default: false
  provisioner: cinder.csi.openstack.org
  volumeBindingMode: WaitForFirstConsumer
  parameters:
    type: storage_premium_perf0   # used in all regions without an entry below
  regions:
  - name: europe-west
    parameters:
      type: premium               # used only in this region
  - name: asia
    unavailable: true             # this storageClass is not created in this region
```
Fixes #1427

**Release note**:

```feature operator
The `storageClasses` in the openstack CloudProfile `providerConfig` now support an optional `regions` list to override `parameters` per region, or to mark a storageClass as `unavailable` in a region.
```

## Summary by CodeRabbit

* **New Features**
  * Added region-specific storage class configuration for OpenStack cloud profiles.
  * Storage classes can now define regional parameter overrides or be marked unavailable in specific regions.
  * Control plane configurations now automatically use only storage classes applicable to their region.

* **Validation**
  * Added checks for valid, unique region names and compatible regional settings.

* **Documentation**
  * Updated configuration references and examples to describe regional storage class settings.


Jonher937 (approved):
We're also in need of this feature. We have a similar problem as yourself providing different storage tiers in different datacenters.

Code changes LGTM.

murwall:
This would be a very good contribution. 
Code looks ok to me but I have no /lgtm power.

PR #1436, opened 2026-09-16
Fixes incorrect reconciliation of wildcard DNS records

### LINKED ISSUE #1434: Incorrect behavior reconciling wildcard records

/area networking
/kind bug
/platform openstack

**What happened**:

We needed the provider to reconcile the wildcard DNS record for our virtual garden. This never happened, even when explicitly adding the reconcile annotation. It will always say that the reconciliation was successful, even when it did not do anything.

The DNS record should have been created in designate.

Create a DNSRecord with a wildcard. You also need to have records that have this subdomain as a suffix, just without the `*`. The records data needs to match as well.

```
*.sub.domain.tld.     A 1.1.1.1
abc.sub.domain.tld. A 1.1.1.1
def.sub.domain.tld. A 1.1.1.1
```

If you now delete the wildcard entry, the reconciliation of its DNSRecords will be successful and not re-create it.

The reason behind it is that any wildcard search for recordsets replaces `*` with a `%`, which will internally look up entries with a LIKE => `name LIKE '%.sub.domain.tld.'`, thus returning anything with the suffix `.sub.domain.tld.`. In `pkg/openstack/client/dns.go` we just return `rss[0]`, resulting in success if the records entry matches.

https://bugs.launchpad.net/designate/+bug/2167327

**Environment**:

- Gardener version (if relevant): 1.149.4
- Extension version: v1.59.0-dev
- Designate version: 2026.1

hebelsan:

/area robustness
/kind bug
/platform openstack

When looking up an existing recordset, getRecordSet passed the requested name to Designate as a server-side filter and blindly returned the first result (rss[0]).
For wildcard names this is unsafe: 
Designate converts the * into a SQL % and runs a LIKE query (upstream bug designate #2167327 https://bugs.launchpad.net/designate/+bug/2167327), so a query for *.sub.domain.tld. also matches every sibling record sharing that suffix (abc.sub.domain.tld., def.sub.domain.tld., …).
  
This caused two failures:
- Create/update: if a sibling with matching records was returned first, reflect.DeepEqual passed and the reconciler reported success without ever creating the wildcard recordset.
- Delete: the wrong (sibling) recordset could be deleted instead of the wildcard.

In both cases the reconciliation reported success while doing nothing / the wrong thing.

Fix: getRecordSet no longer trusts the server-side filter.
It iterates the returned recordsets and returns only the one whose name matches exactly, normalizing both sides (trailing dot + \052→* escape) so it works whether Designate returns *.sub.domain.tld. or \052.sub.domain.tld..

Adds unit tests covering exact-match selection over siblings.

Fixes https://github.com/gardener/gardener-extension-provider-openstack/issues/1434

**Release note**:

```bugfix operator
Fixes incorrect reconciliation of wildcard DNS records
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNS recordset lookup to return the exact requested name, including wildcard recordsets, instead of potentially selecting a similarly named recordset.
  * Correctly handles cases where no exact matching recordset is returned.

* **Tests**
  * Added coverage for exact matches, wildcard names, sibling recordsets, and empty results.


hebelsan:
Sorry @modzilla99 I just now saw that you already have opened a fix PR for the issue: https://github.com/gardener/gardener-extension-provider-openstack/pull/1435.
If you're okay with it, I'd keep this PR as is, since it includes tests and is more forgiving when multiple recordsets with the same name are found, it doesn't return an error in that case. 
I'd also add you as a co-author.

